PR #1, opened 2011-11-22
Publisher and Subscriber Prototype

dstuebe:
Added non functional skeleton of publisher and subscriber prototypes for discussion with the architecture & COI team.

### DIFF
--- a/pyon/net/endpoint.py
+++ b/pyon/net/endpoint.py
@@ -359,6 +359,46 @@ class Publisher(EndpointFactory):
         if self._pub_ep:
             self._pub_ep.close()
 
+class PrototypePublisher(Publisher):
+    """
+    High level abstraction of EndPoint for publishing messages to a stream
+    """
+
+    def __init__(self, exchange_name, credential, **kwargs):
+        """
+        init method used by a publisher factory
+        """
+        self._exchange_name = exchange_name
+        self._credential = credential
+
+        # @TODO Magic to make the endpoint factory do what we want - have a name and publish TO credential
+
+class PrototypePublisherFactory(object):
+    """
+    A factory which creates an end point on initialization and then uses that end point in many publishers for different
+    streams
+    """
+
+    def __init__(self, exchange_name=''):
+        """
+        Call EMS (container for now) to make an end point in the exchange which will be used to publish to streams
+        """
+
+        self._exchange_name = exchange_name # could also be an anonymous name...
+
+        # @TODO create an endpoint - what kind? It should not have any binding - it is a producer only
+
+    def create_publisher(self, streamid):
+        """
+        Call pubsub service to register this exchange name (endpoint) to publish on a particular stream
+        """
+
+        # @TODO make a pubsub service client...
+        credential = clients.pubsub.register_producer(self._exchange_name, streamid)
+
+        return PrototypePublisher(ep=self, credential=credential)
+
+
 class SubscriberEndpoint(Endpoint):
     """
     @TODO: Should have routing mechanics, possibly shared with other listener endpoint types
@@ -396,6 +436,40 @@ class Subscriber(ListeningEndpointFactory):
     def create_endpoint(self, **kwargs):
         log.debug("Subscriber.create_endpoint override")
         return ListeningEndpointFactory.create_endpoint(self, callback=self._callback, **kwargs)
+
+
+class PrototypeSubscriber(Subscriber):
+
+    def __init__(self, exchange_name, **kwargs):
+
+        self._exchange_name = exchange_name
+
+        Subscriber.__init__(self, **kwargs)
+        # Queue creation happens here
+
+    def set_callback(self, callback):
+
+        self._callback = callback
+
+
+    def subscribe(self, subscription):
+
+        subscription.exchange_name=self._exchange_name # the name of the subscriber to bind the routing rule to
+        res = clients.pubsub.subscribe(subscription)
+        # The binding happens in the pubsub service
+
+
+    def activate_subscriber(self):
+        """
+        Start consuming from the queue
+        """
+
+    def deactivate_subscriber(self):
+        """
+        Stop consuming from the queue
+        """
+
+
 
 #
 # BIDIRECTIONAL ENDPOINTS


pyon/net/endpoint.py
```diff
@@ -359,6 +359,46 @@ class Publisher(EndpointFactory):
...
+    """
+
+    def __init__(self, exchange_name, credential, **kwargs):
+        """
+        init method used by a publisher factory
+        """
+        self._exchange_name = exchange_name
+        self._credential = credential
+
+        # @TODO Magic to make the endpoint factory do what we want - have a name and publish TO credential
```
dstuebe, line 374: Inherits from Publisher, but allows the pubsub service to determine the TO for publishing a message.

pyon/net/endpoint.py
```diff
@@ -359,6 +359,46 @@ class Publisher(EndpointFactory):
...
+
+class PrototypePublisherFactory(object):
```
dstuebe, line 376: All the service calls to EMS and PubSub should go in the is class - the Publisher is just an object that knows what it is Publishing TO and how to publish.

pyon/net/endpoint.py
```diff
@@ -397,6 +437,40 @@ class Subscriber(ListeningEndpointFactory):
         log.debug("Subscriber.create_endpoint override")
         return ListeningEndpointFactory.create_endpoint(self, callback=self._callback, **kwargs)
 
+
+class PrototypeSubscriber(Subscriber):
```
dstuebe, line 441: This is the object that is actually subscribed. It has an exchange name (could be anonymous/random) and provides the ability to set the callback when a message is received. It allows for many subscriptions using the same exchange name where messages are received on a FIFO queue.

pyon/net/endpoint.py
```diff
@@ -397,6 +437,40 @@ class Subscriber(ListeningEndpointFactory):
...
+
+    def __init__(self, exchange_name, **kwargs):
+
+        self._exchange_name = exchange_name
+
+        Subscriber.__init__(self, **kwargs)
+        # Queue creation happens here
```
dstuebe, line 448: The subscriber owns and creates the queue where messages are delivered.

pyon/net/endpoint.py
```diff
@@ -397,6 +437,40 @@ class Subscriber(ListeningEndpointFactory):
...
+
+
+    def subscribe(self, subscription):
+
+        subscription.exchange_name=self._exchange_name # the name of the subscriber to bind the routing rule to
+        res = clients.pubsub.subscribe(subscription)
+        # The binding happens in the pubsub service
+
+
+    def activate_subscriber(self):
```
dstuebe, line 462: Since it is the subscriber object that has the queue - it is the subscriber that can activate and deactivate the delivery of a subscription. Not the pubsub service. The pubsub service could only turn on and off the binding which is not the behavior that we want.

rajvikram:
This is the test code for netCDF4-python

---

from netCDF4 import Dataset
from datetime import datetime, timedelta
from netCDF4 import num2date, date2num
import gviz_api
import time

def readNetCDF():

```
# using NetCDF4    library to read up a .nc file
ncFile =
```

Dataset('/var/www/map_viz/data/CGSN_AS02CPSM/AS02CPSM_sbe16.nc', 'r',
format='NETCDF4')
    print "\n netcdf format = " + ncFile.file_format + "\n"
    varList = ncFile.variables.keys()

```
# For creating the Json object we will do a redundant operation where
```

the NetCDF is loaded up in
    # a google visualization API dataTable and then the ToJSonResponse
function is used to create the
    # response object
    dataDescription = [('time','datetime', 'time')]
    dataContent = []

```
# populate the dataDescription dictionary with the tuples describing
```

the columns of the data table
    # Use the name of the variables in the netCDF file as keys for the
dictionary
    for i in varList:
        # ignore some of the variables since not everything applies to the
graphs
        if  i == 'time' or i == 'longitude' or i == 'latitude':
            continue
        dataDescription[len(dataDescription):] = [(str(i),'number', str(i))]

```
# extract time and convert to a standard format
timeLine = ncFile.variables['time']
timeList = num2date(timeLine[:], 'days since 1950-01-01T00:00:00Z',
```

'standard')

```
cndcList = ncFile.variables['CNDC']

#dataContent = list(zip(timeList, cndcList))

#print timeLine[0]
#print timeList[0]
#print cndcList[0][0]


tic = time.time()
for i in range(len(timeList)):
    dataContent[len(dataContent):] = [timeList, str(cndcList[0][i])]

toc = time.time() - tic

print 'Time taken : ' + str(toc)
#print dataContent

ncFile.close()
```

if **name** == '**main**':
    readNetCDF()

On Tue, Nov 22, 2011 at 7:08 AM, David Stuebe <
reply@reply.github.com

pyon/net/endpoint.py
dstuebe, line 376: Change to PublisherRegistrar

pyon/net/endpoint.py
dstuebe, line 441: Separate into StreamSubscriber and subscriberRegistrar
Make sure to allow for creating a new exchange_name or using an existing one.

pyon/net/endpoint.py
dstuebe, line 462: Add methods for attach and detach to the subscriber. Activate and Deactivate stay withe the subscriber registrar.

pyon/net/endpoint.py
```diff
@@ -359,6 +359,46 @@ class Publisher(EndpointFactory):
         if self._pub_ep:
             self._pub_ep.close()
 
+class PrototypePublisher(Publisher):
+    """
+    High level abstraction of EndPoint for publishing messages to a stream
+    """
+
+    def __init__(self, exchange_name, credential, **kwargs):
```
dstuebe, line 367: Change credential to stream_route

pyon/net/endpoint.py
```diff
@@ -359,6 +359,46 @@ class Publisher(EndpointFactory):
         if self._pub_ep:
             self._pub_ep.close()
 
+class PrototypePublisher(Publisher):
```
dstuebe, line 362: Change to StreamPublisher

dstuebe:
Closed this pull request - submitting a new one with updates.

PR #2, opened 2011-11-29
Pubsubobj

dstuebe:
Updated the publisher and subscriber and added the registrar objects. The interface to these objects is now in place but I need help from the COI team with the endpoint layer. It is unclear how to use a process to get a FROM exchange name and it not currently possible to set that for the publisher. I think there is still some work to be done in the endpoint layer to make all this work, but hopefully seeing how it will be connected for the DM pubsub layer will help inform the effort.

Signed-off-by: David Stuebe dstuebe@asascience.com


mmeisinger:
Ok, manually merged this pull request, now in master

PR #3, opened 2011-11-29
Refactored generate_interfaces.py to be less spaghetti-based

ianfixes:
Overview of changes:
1. Converted to a class instead of using one big function
2. Got rid of OS-specific path splitting and joining
3. Converted to use class-level constants for directory names
4. Some irrelevant code removed
5. Comments added

Tested -- produces the same output as the original (with the exception of timestamps, for obvious reasons).

The initial goal of this edit was to support local overrides to extern/ion-definitions; instead of making changes to that directory or working entirely out of a possibly-obsolete personal fork, the idea would be to have a separate local-definitions directory whose contents would override the corresponding file in ion-definitions.  Said functionality has not been added, but the groundwork is certainly there (via the yml_directory_walk function).  (This feature would require some coordination with the way that object definitions are handled.)

As-is, this pull request is for a cleaned-up version of the original.

### DIFF
--- a/scripts/generate_interfaces.py
+++ b/scripts/generate_interfaces.py
@@ -19,15 +19,21 @@ import argparse
 from pyon.service.service import BaseService
 from pyon.util.containers import named_any
 
-# Do not remove any of the imports below this comment!!!!!!
+# Adam Smith: Do not remove any of the imports below this comment!!!!!!
 from pyon.core.object import IonYamlLoader, service_name_from_file_name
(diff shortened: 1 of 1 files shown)


daf:
(speaking with Ian now via Skype, but for others)

I'm just finishing my own refactor of the script to support client generation which changes enough to make this very difficult to merge just now.  I will manually merge some of the contents of this PR then, but for now I must close.

PR #4, opened 2011-12-08
Switch back to https URL for submodules.

oldpatricka:
You can't do an anonymous checkout with the ssh protocol url, which
makes it difficult to do a deployment without passing along github
credentials.

I realize this can be a bit annoying, since git will ask you for your username/password each time you want to push, but you can avoid this by creating a .netrc file like the following:

```
$ cat ~/.netrc:
machine github.com
login yourusername
password secretpassword
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "extern/ion-definitions"]
 	path = extern/ion-definitions
-	url = git@github.com:ooici/ion-definitions.git
+	url = https://github.com/ooici/ion-definitions.git


daf:
Seems like the right approach.  I currently get asked for my username/password anyway when working in `extern/ion-definitions` so it's no more trouble currently..

Related, have you tested what this change means for existing developers?  Will it require them to do anything to their submodules?  Will the hook scripts that most are (should be!) using just pick it up no problem?

oldpatricka:
Hmm, I'm not sure about the hook scripts, since I've never used them, but I noticed that this url was https prior to bb964e7f7e48086d4e88778f51e9807db22abf5f on November 1, so it has happened before. Developers may have to run a "git submodule sync", to switch to https once this commit is in their tree, but I don't think it'll be necessary, since their existing clones should work just fine with the old ssh protocol URL.

mmeisinger:
Concur, this is probably the right approach.
Background: The GIT url affects this file at the time the submodule .git directory is created:
  extern/ion-definitions/.git/config
It contains the actual URL used. It can be changed manually to avoid the ~/.netrc trick

mmeisinger:
Also, two more comments:
1. the same needs to be applied to other repos with submodules (e.g. coi-services)
2. For developers that already have created their repos and submodules, they will keep the current contents of extern/ion-definitions/.git/config I suppose

PR #6, opened 2011-12-13
Channel refactor

daf:
Finally.

This is a big refactor based on conversations with @mmeisinger which is a simplification of the channel layer in the messaging stack.

69 unit tests over channel and endpoint layers.

Some pieces are still missing but can be implemented easily now that all this in place:
- request/response anon queue pooling is disabled currently for correctness at the cost of raw msging speed
- no funneling into an "ion process"
- in memory messaging

To be reviewed by Michael before merging please.


daf:
Spoke with Michael on a tour of the code - need to add a few more things:
- RecvChannel close_impl is too agressive in trying to delete queues - can delete a service queue out from underneath a second listener
- There should be a destroy_listener() pairing to setup_listener() that does the destroy (hard to tell when to call it)
- Reject should be exposed as well as ack.

Working on these, will merge after.

PR #10, opened 2011-12-16
RPC messages send IonObject, op in headers instead of dict

daf:
TBR @mmeisinger

HUGE PROVISIONS:
- Container must add to obj_registry on startup to make sure it has IContainerAgent related *_in messages (pycc and control_cc require this)
- object loading in pyon/core/object MUST have the OrderedDict yaml patch (so we can map args to kwargs in calls)
- RPCClient reworked to better support dynamically defined interfaces, IonObject based routing, old routing removed
- generate_interfaces support for new rpc style calls, always --force now (we keep changing it too much)

### DIFF
--- a/pyon/container/cc.py
+++ b/pyon/container/cc.py
@@ -25,17 +25,16 @@ import msgpack
 import atexit
 import signal
 
-from pyon.core.bootstrap import CFG, sys_name, bootstrap_pyon
+from pyon.core.bootstrap import CFG, sys_name, bootstrap_pyon, obj_registry
 
 from pyon.container.apps import AppManager
 from pyon.container.procs import ProcManager
 from pyon.directory.directory import Directory
-from pyon.net.endpoint import ProcessRPCServer, RPCServer
+from pyon.net.endpoint import ProcessRPCServer, RPCServer, RPCClient
 from pyon.net import messaging
 from pyon.util.log import log
 from pyon.util.containers import DictModifier, dict_merge
 from pyon.ion.exchange import ExchangeManager
-
 
 class IContainerAgent(Interface):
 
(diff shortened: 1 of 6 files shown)


shenrie:
Hi Dave,

I am seeing a situation related to the changes you recently made for the
keyword arguments, ie. AssertionError: You MUST used named keyword args when
calling a dynamically generated remote method.   Perhaps you could run the
following scenario and let me know if there is something I am doing wrong or
if there is actually a bug somewhere in the endpoint code.

First some background, there are a couple helper functions in the service
gateway service which are basically wrapper functions for working with the
bank client so that it can be called from an HTTP request. The bank_complete
rel file has been modified to start the service gateway service as well;
which provides a restful HTTP interface to the AMQP service calls. What I am
doing used to work, so I am not sure I understand what has changed to cause
the problems I am seeing.
- check out the latest pyon and coi-services code
- start the container: bin/pycc --rel=res/deploy/examples/bank_complete.yml
- first, see that you can run the bank client multiple times without
  failure:
  from examples.bank.bank_client import run_client
  run_client(cc)
  run_client(cc)
- then try doing it using the wrapper function, which basically calls the
  same run_client method but then sets up an RPC process endpoint to list out
  the accounts which where created:
  from ion.services.coi.service_gateway_service import create_accounts
  create_accounts()
  create_accounts()
- Notice that it fails on the second call and hangs the container interface
  with the error:
  File "/Users/sphenrie/Documents/Dev/code/pyon/pyon/net/endpoint.py", line
  613, in svcmethod
  assert len(args)==0, "You MUST used named keyword args when calling a
  dynamically generated remote method"      # we have no way of getting
  correct order
  AssertionError: You MUST used named keyword args when calling a dynamically
  generated remote method

So the question is why does it work the first time but not with subsequent
calls?   

Let me know if you have any questions duplicating this scenario.

Thanks,
Stephen

---

System Architect

PR #15, opened 2012-01-13
Spawn proc

dstuebe:
Fixed spawn process interface so that it returns a process id.


dstuebe:
Merging to get this over with...

PR #16, opened 2012-01-13
Interface fix

dstuebe:


dstuebe:
Never mind - used a push...

PR #18, opened 2012-01-18
Class generation

thomaslennan:
These are the complete code changes necessary to support usage of data model classes generated from the yaml.


daf:
This is a good start, but is waiting for a few more, and requires a separate PR from ion-definitions to work properly.  I will merge when we deem this is ready to go.

PR #19, opened 2012-01-18
Stream Example Patch

lukecampbell:


examples/stream/stream_consumer.py
```diff
@@ -31,9 +31,15 @@ class StreamConsumer(StreamProcess):
...
+        if stream_route:
+            self.channel = self.container.node.channel(BindingChannel)
+            self.channel.setup_listener(('science_data',stream_route),binding='stream_example')
 
     def on_quit(self):
         log.debug("StreamConsumer quit")
 
     def process(self, packet):
-        log.debug("Processing: %s", packet)
+        log.debug('(%s): Received Packet' % self.name )
```
daf, line 44: Future ref, use the `log.debug("bla bla %s", self.name)` syntax - the % operator is unnecessary and is slower

PR #20, opened 2012-01-19
Class generation final

thomaslennan:
Pyon changes for supporting class generation from yaml.  This is a pull request for the COI Team only.  Guys, I'm hoping this is the final changes required to get class generation in place.  Per Dave's request, I'm now spitting out the named args in the **init** for each class.  I'm hoping we can push this ASAP so that the definitions and impls don't drift much. Please have a look ASAP and give it a thumbs up/down.

### DIFF
--- a/pyon/agent/agent.py
+++ b/pyon/agent/agent.py
@@ -33,13 +33,13 @@ class ResourceAgent(BaseResourceAgent):
     def _on_quit(self):
         self.clients.directory.unregister("/Agents", self.id)
 
-    def negotiate(self, sap_in={}):
+    def negotiate(self, sap_in=None):
         pass
 
-    def execute(self, command={}):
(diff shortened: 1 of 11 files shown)


mmeisinger:
This is an impressive pull request. It looks all good and works for me. I'll manually merge in a bit after some more testing.

The only issue that I have is that generate_interfaces is now even more spaghetti code. I would like the object parsing functions actually residing in a pyon.core module, imported and called modularly from generate_interfaces.

thomaslennan:
Agreed.  This can be done as part of the cleanup of the parsing logic
to use a real parser vs massive nested if/else clauses.

On Thu, Jan 19, 2012 at 4:48 PM, Michael Meisinger
reply@reply.github.com
wrote:

PR #21, opened 2012-01-20
Stream example refactor

dstuebe:
Michael should review and close or merge this PR. It is part of an ongoing discussion about the stream example.

### DIFF
--- a/examples/stream/stream_binder.py
+++ b/examples/stream/stream_binder.py
@@ -1,0 +1,51 @@
+#!/usr/bin/env python
+
+"""
+@author David Stuebe
+
+@brief This process exists to take the place of the pubsub service which implements the subscription and create the
+rabbitmq binding because the pubsub service is not running in the simple stream example.
+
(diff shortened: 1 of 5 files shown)


mmeisinger:
Merged it manually. All fine, some comments apply. Thanks!

I would have liked the stream_consumer process be spawned from inside the binder, with a route (=queue name) given in the config. Keeping the consumer in the stream.yml file does not make this possible, because you have to hardcode the queue name.

dstuebe:
Not sure I understand what you are after here - lets talk tomorrow.

mmeisinger:
Let's talk then.

I think I was expecting more real scenario type "subscription" and binding, with the binder process spawning the consumer, like in a typical transform scenario.

I think the example benefits from the simplicity it is now, with the hardcoded queue name and the explicit consume process start.

The real scenario will provide the pubsub/transform example I was looking for. 

So things are good and we have great progress!

Thanks,
Michael

PR #23, opened 2012-01-25
Integration test

thomaslennan:
Huge pull request that includes implementation of integration tests for the resource registry, directory and data store services.  Ran across a wide variety of issues and holes as I developed the tests.  This pull request needs to be completed in conjunction with the pull requests for ion-definitions and coi-services.

### DIFF
--- a/pyon/core/exception.py
+++ b/pyon/core/exception.py
@@ -100,3 +100,11 @@ class ContainerAppError(ContainerError):
     '''
     '''
     status_code = 554
+
+exception_map = {}
+import inspect
+import sys
+for name, obj in inspect.getmembers(sys.modules[__name__]):
+    if inspect.isclass(obj):
+        if hasattr(obj, "status_code"):
+            exception_map[str(obj.status_code)] = obj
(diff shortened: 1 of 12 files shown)


pyon/core/exception.py
```diff
@@ -100,3 +100,11 @@ class ContainerAppError(ContainerError):
     '''
     '''
     status_code = 554
+
+exception_map = {}
+import inspect
+import sys
+for name, obj in inspect.getmembers(sys.modules[__name__]):
```
daf, line 107: This is rather unnecessary - for a list of exceptions that don't change often, we should pretty much just statically define that list.

pyon/core/object.py
```diff
@@ -85,6 +85,10 @@ def _validate(self):
     def __contains__(self, item):
         return hasattr(self, item)
 
+    def __setattr__(self, name, value):
+        if name not in self._schema and name != "_id" and name != "_rev":
+            raise AttributeError("'%s' object has no attribute '%s'" % (type(self).__name__, name))
+        self.__dict__[name] = value
```
daf, line 91: Should this be calling object's **setattr** instead?  (is that possible to do?)

PR #27, opened 2012-01-27
Numpy Interceptor

lukecampbell:
A numpy codec was added to the interceptor stack to support sending and
receiving numpy arrays. The data translation has been confirmed as
lossless at various precisions (including float32). If the recipient
does not have numpy it is simply a dictionary packed with strings, so it
will not break a client receiving a numpy object.


daf:
good2rock

PR #28, opened 2012-01-31
Luke: Datastore patch for view templates

lukecampbell:
Datastore implementation fix. 
View templates are now based on datastore_name not a generic one size fits all template.


lukecampbell:
Luke: In light of recent discussions, I have opted to extend the existing class in lieu of changing it.

PR #30, opened 2012-02-03
Fail fast support in container

daf:
Changes to support the container's fail fast, also touches/cleans up process manager stuff.

TBR @mmeisinger
- ensure_ready: instead of waiting for RPCServer specific ready event, make the ready event generic to any spawned process, will also trap errors on startup resulting in greenlet deaths and raise an exception in the waiting greenlet.  Cleans up startup logic.
- Monitor for failed greenlets/procs spawned by proc manager/proc supervisor.
- GreenIonProcess now has an initializer instead of setting instance vars in the target method
- Channel layer now uses custom synchronous wrapper instead of generic blocking_cb, can handle errors

### DIFF
--- a/pyon/container/cc.py
+++ b/pyon/container/cc.py
@@ -4,7 +4,7 @@
 Capability Container base class
 """
 
-__author__ = 'Adam R. Smith, Michael Meisinger'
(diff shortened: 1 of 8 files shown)


mmeisinger:
Looks good and definitely the right way in robustifying the container. The PyonProcess/supervisor layer may need a bit of refactoring in the future to make it less confusing with the ION process layer.

PR #31, opened 2012-02-03
Blog

dstuebe:
Dm integration work to implement elaboration capabilities using simple blog data.


mmeisinger:
That's ok. Merge if all works. 
We might want to discuss harmonization of the datastore layer and extension on the design level shortly (IndexMgmtService etc, replicated datastore etc)

dstuebe:
Yes - very much want to talk about that, but did not want to make changes without coordination.

PR #33, opened 2012-02-07
Luke: (Terminate process) Quick Patch

lukecampbell:
- Removed the assertion, we now have processes without subscriber
  endpoints (simple process), the assertion may fail in the future.


mmeisinger:
Ok, good. Although Dave F probably had an intent with this assertion

PR #37, opened 2012-02-08
Fixed issues in Directory and Datastore

ianfixes:
Datastore: incorrect init argument being supplied to mockdb, removed

Directory: if __instance falls into the state "NEW", you're screwed.  Caught and fixed, with a warning.


daf:
This will work for now, but for later:
- let's expand MockDB to actually take that profile kwarg
- more try/except/reporting of problems creating Directory object.  Likely problems with couchdb connectivity.

PR #39, opened 2012-02-08
Remove singletons from DatastoreManager and related classes

daf:
These cause problems, especially between integration test runs.  Global state should be as isolated as possible - in our case, the Container instance is the one point for all of that.
- StateRepository
- EventRepository
- Directory

Only the EventRepository change has logical implications - EventPublishers must be constructed with a given EventRepository to gain functionality.  It's not perfect yet but better than having deps on a Container instance (which would have to late bind etc).  Can probably make a factory style pattern in the Container to create rigged up EventPublishers, or a more derived version of them that knows how to pull the correct event repo.

### DIFF
--- a/pyon/container/cc.py
+++ b/pyon/container/cc.py
@@ -123,7 +123,7 @@ class Container(BaseContainerAgent):
         self.app_manager.start()
 
         # Create other repositories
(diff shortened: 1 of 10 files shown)


mmeisinger:
Manually merged

PR #40, opened 2012-02-09
Luke: Datastore

lukecampbell:
Added an import to couchdb datastore, necessary for ingestion to function correctly.


mmeisinger:
ok

PR #41, opened 2012-02-09
Clear couch

dstuebe:
Added a command line utility which is installed in the bin directory by buildout. The utility allows the user the clear all databases in couch matching a certain prefix such as 'ion_sci053mac001_local'.

### DIFF
--- a/buildout.cfg
+++ b/buildout.cfg
@@ -62,6 +62,7 @@ scripts =
     generate_interfaces
     ipython
     json_report
+    clear_couch
 entry-points =
     ipython=scripts.ipython:main
     nosetests=nose:run_exit
--- a/pyon/datastore/clear_couch_util.py
+++ b/pyon/datastore/clear_couch_util.py
@@ -64,7 +64,7 @@ def main():
 
     for db_name in db_list:
 
-        if db_name.find(prefix) != -1:
+        if db_name.startswith(prefix):
 
             db_server.delete_datastore(db_name)
             print '$ Cleared couch db named: %s' % db_name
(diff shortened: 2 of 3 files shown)


daf:
A few possible things: 
- Can we have the ability to use this script to list all known databases?
- Can we use the same defaults to construct the prefix/sysname as pyon does?
- Can we get a "I'm going to delete ALL THESE DATABASES: continue?"

dstuebe:
re: A few possible things:
• Can we have the ability to use this script to list all known databases?
•• I don't think we need list in the command line tool - this should not replace or augment the couch web interface

• Can we use the same defaults to construct the prefix/sysname as pyon does?
•• I am specifically not using sysname as the argument. This is a couch tool, not a pyon tool so it does not know about sysnames or how to construct the name scoping.

• Can we get a "I'm going to delete ALL THESE DATABASES: continue?"
•• We can make it interactive. Do you think it is worth adding it now?

daf:
My goal with the suggestions is to make it easier for the end user to accomplish the goal, which is to clear couch databases.  How does the user remember which couch databases to clear (aka which prefix)?  Would not the majority of things I want to clear be the automatic databases created on pycc startup, which by default is scoped to the default sysname prefix?

dstuebe:
$ bin/clear_couch ion
Done...

daf:
heh, ok.

mmeisinger:
I would like to see this as one bit in the larger integration strategy. I don't want to rush in partial tools addressing symptoms. I will have a look at the overall strategy and discuss with Jamie and Tim about how to proceed.

mmeisinger:
This tool as a standalone tool is reasonable. I would like to see it as part of a system solution, not a standalone tool though. This just leads to tool sprawl and confusion, especially if replicated identically in coi-services etc.

I'm merging this to avoid possible blocks now, but we may change this soon.

If avoidable, do not add to coi-services. It's ok to run this from pyon. I will find a better solution for system runs, e.g. within rel files.

Don't push this issue too much anymore. However, let me know any suggestions.

PR #42, opened 2012-02-09
Dm datastore

lukecampbell:
Added several views to the scidata profile.


dstuebe:
Merging update to couch views

PR #43, opened 2012-02-10
Luke: (Datastore) Patched bug with reduce queries.

lukecampbell:


dstuebe:
Hot patch - looks good!

PR #46, opened 2012-02-13
Luke: (Datastore) (ctd_data_stream)

lukecampbell:
- Added conditional to create datastream for hdf file creation.
- Added full map/reduce for dynamic bounds checking.


dstuebe:
Hi Luke
Looks good if this keeps you moving.

David

mmeisinger:
ok.

regarding the map/reduce couch functions: if possible, can they be included readable in correct flow instead of using backslash-n inline newlines?

Famous quote: "code is to be read by humans and only occasionally executed by machines"

We can split the one config file into several if it gets too long. We can make a separate file out of every design, whatever works.

lukecampbell:
I don't format it that way, couch does. I copy it from couch.
I keep a gist offline of the reduces.

lukecampbell:
https://gist.github.com/1781675

lukecampbell:
If you want, I can move into another file and run it through a beautifier so it looks clean.

mmeisinger:
No it's fine, if you include the link to the gist and keep the gist up to date.

There is great stuff in these functions and it would be good to keep it readable.

I was not aware that you edit elsewhere and then copy/paste over.

On a related note: if you have good CouchDB tools for everything involved to install, pls create a little how-to page.

Thanks!

PR #54, opened 2012-02-14
Luke: Added and changed maps

lukecampbell:
- Added a map copy of posts_join_comments to dataset_by_id to the posts
  profile
- Added a map stream_join_granule and dataset_by_id to datasets profile.

In the future, all datasets, posts, science data etc will have a default
map whereby the key is a tuple consisting of the key identifier followed
by a type identifier usually an integer. This allows maps and reduces to
be executed quickly. Performance gains from this pattern are an
absolute must, the nomenclature dataset_by_id is just a logical
compliment to this consequence of events.

### DIFF
--- a/pyon/datastore/couchdb/couchdb_config.py
+++ b/pyon/datastore/couchdb/couchdb_config.py
@@ -200,6 +200,9 @@ function(doc) {
         },
     },
     'posts' : {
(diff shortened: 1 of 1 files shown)


dstuebe:
Map reduce functions are looking good!

PR #57, opened 2012-02-18
Hdf test

dstuebe:
Pull request to add hdf as as dependency.
Nothing will fail without it - yet.


dstuebe:
Let Jamie merge this when she is ready... Not late Friday afternoon!

mmeisinger:
looks good. go when Jamie is ready

PR #58, opened 2012-02-21
To ooici

dstuebe:
Change Summary:
Added File System tool for relative path management
Cleaned up HdfCodec and associated tests using File System tool
Added sha1 to encoding object in the ctd_stream data prototype object.


dstuebe:
Working to add recursive method to make directories more flexible.

PR #61, opened 2012-02-27
Adding Constructor Deconstructor interfaces for working with granules

dstuebe:


mmeisinger:
Looks good on first glance.

Can we rename Deconstructor to something better? I don't like this term at all.
How about StreamReceiver, StreamParser, DatasetBuilder, etc ... or something better fitting

daf:
Merging due to coi-services having a merge referencing code in this.  DavidS will fix the naming later.

PR #63, opened 2012-02-28
Dupe association checking

thomaslennan:
Michael, Stephen, the pyon changes related to dupe association checking.


pyon/datastore/couchdb/couchdb_datastore.py
```diff
@@ -503,6 +503,8 @@ def find_associations(self, subject=None, predicate=None, obj=None, id_only=True
             key = [subject_id, object_id]
             if predicate:
                 key.append(predicate)
+            if assoc_type:
```
mmeisinger, line 506: The challenge here is that if predicate is not present, you still need to add it to the query. This means assoc_type can only be given with a predicate. This is good, but please check this condition.

mmeisinger:
looks good to me. please merge when ready

PR #64, opened 2012-02-28
Added support for syslog logging

thomaslennan:
Implemented class to override built-in syslog handler since it drops log records that exceed the transport layer's MTU.  My class chunks long log records into sub-MTU size log records to at least ensure the log information is recorded.


mmeisinger:
very good, thanks. merge when ready

PR #65, opened 2012-02-29
Generate random sysname on testing scenarios

daf:
Not sure if this is the correct approach, but seems to work.  Should limit the whole broker-cycling thing.


mmeisinger:
Looks good.
I like the use of get_sys_name() for this distinction.
Can we make the direct use of sys_name prohibitive in imports somehow?

PR #68, opened 2012-03-02
Split Int tests force_clean into two distinct paths: create/delete

daf:


mmeisinger:
looks good pls merge if Jamie's ok

shenrie:
Hi Dave,

I could use your help in setting up an event. Unfortunately my code is not
yet checked in, so I can't show you exactly what I am doing, but I am hoping
you can give me some help in the right direction.

I need to be able to notify containers to reload their policy for a service
( or generic resource ) whenever the policy for a service/resource changes.
I have a singleton class which is checked in called governance_controller in
the container, though I am moving its startup to the container startup class
with my next checkin,  and I would like for it to be able for it to receive
events that are published by the policy management service when policy
objects changed. The governance_controller in each container should be able
to register some kind of call back function that would get called when an
event Is published ( that contains the name of the service/resource ), so
that the callback can then call the policy management service to request the
set of polices for the service/resource that was in the event.

Make sense?

Any directional help you can provide would be greatly appreciated.

Thanks
Stephen

---

System Architect

PR #70, opened 2012-03-09
SFlow integration (counter stats, RPC transactions)

daf:
Purely opt-in.  Read the `pyon/util/sflow.py` file for config instructions.  Needs a special hsflowd instance running somewhere.

### DIFF
--- a/pyon/container/cc.py
+++ b/pyon/container/cc.py
@@ -169,7 +169,7 @@ class Container(BaseContainerAgent):
         self.governance_controller.start()
         self._capabilities.append("GOVERNANCE_CONTROLLER")
 
-        if CFG.container.get('sflow', False):
+        if CFG.container.get('sflow', {}).get('enabled', False):
             self.sflow_manager.start()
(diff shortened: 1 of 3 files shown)


mmeisinger:
Very nice solution. Thanks again. The one thing to improve is to not call the classes SFlowManager, but something like stats_manager. Then define one implementation of this for sFlow.I don't want the choice of tool leak into the other code. The fact that we are emitting stats is fine. Maybe we can refactor this past LCA.

daf:
The intent is to refactor this class to be derived from a more general
implementation as you note. This can take place post lca.
On Mar 13, 2012 12:43 PM, "Michael Meisinger" <
reply@reply.github.com>
wrote:

mmeisinger:
The PR broke the pyon UNIT and INT tests. Had to revert this. Please check if any recent changes to pyon endpoint break this PR.

daf:
Will check when i am home tonight, thanks
On Mar 13, 2012 3:13 PM, "Michael Meisinger" <
reply@reply.github.com>
wrote:

PR #71, opened 2012-03-12
Minor changes

dstuebe:
- Added stream def for USGS data
- Added more tests for HDF array iterator
- Fixed a bug in the couch query logic for dataset bounds


mmeisinger:
ok with me

PR #77, opened 2012-03-27
Luke: create now appends id,rev to obj

lukecampbell:


dstuebe:
This is a suggested change which should allow the developer to create an object and reference its id while still able to add more fields. It avoids an extra resource registry call to read that resource back again.

alt current senario
svc -> res_reg: create(Res)
res_reg --> svc: id, rev
note over svc: Res is no longer in a valid state

svc -> res_reg:read(id)
res_reg --> svc: Res

note over svc: make changes to Res, use its id for associations

svc -> res_reg:update(Res)
res_reg --> svc: id, rev

else pull request change in pattern

svc -> res_reg: create(Res)
res_reg --> svc: id, rev
note over svc: Create method sets id, rev on the input argument Res

note over svc: make changes to Res, use its id for associations

svc -> res_reg:update(Res)
res_reg --> svc: id, rev

end

Basically we avoid an un needed call to read the resource back from the resource registry.

mmeisinger:
I thought about this before long and hard and there is no easy solution. The problem is that the RR is called via a service and so any change to the object is not propagated back. Higher level services should NOT call the RR via its native API and rely on the side effect of an updated object.

In addition, if that were the case, then calling native and via a the RR service would have different behavior. This hidden side effect is what we need to avoid.

I suggest not optimizing at this point and just doing a read again before the second update. This should be a rare case anyways.

Or better, avoid the first create and then update in the first place by rearranging service calls.

lukecampbell:
We found a catch twenty-two with that, we need a transform id to append to the config, we need a process id from CEI to make a transform, and we need a config to launch the process...

This is in a few places as well, TMS is just the example I bring to light.

PR #78, opened 2012-03-28
Luke: Assertion Wrapper

lukecampbell:


mmeisinger:
Thanks this is very good

PR #79, opened 2012-04-03
Transform Benchmarking

lukecampbell:


daf:
We've used this successfully in office to get some load testing going.  Very useful and clean (apart from shutdown, as a testing tool though it's fine)

PR #80, opened 2012-04-16
Channel FSM, accept as a context manager

daf:
Channel made more robust, defining transitions between states.

TBR @mmeisinger


daf:
going to merge, we can revert if issues

mmeisinger:
Sounds good. 

Thanks,
Michael

PR #81, opened 2012-04-20
Msgpack numpy

dstuebe:
Added initial support for numpy arrays encoded using msgpack object hook.
This method is inefficient because it promotes all floating point numbers to double during transfer and specifies the type of each atom in the array.

Future work in COI can make use of the msgpack object hook to avoid walking the object tree in pyon. MsgPack allows you to provide your own method for packing an unpacking object types.

David


daf:
Looks good, thanks!

PR #82, opened 2012-04-26
the assertion for datastore_name crashes the container

lukecampbell:
Assertion replaced with exception raising mechanism which allows clients to catch and handle exceptions as well as it being logged as an issue in lieu of the container dying not so-gracefully.


daf:
This is fine.  For anyone else looking: validateTrue raises an IonException derived class which is non-fatal in a messaging context.

PR #83, opened 2012-04-27
Refactor arg_check

lukecampbell:
Changes the arg_check method wrappers to use PEP8 convention for method names in lieu of `unittest.TestCase` convention.

To make Pyon-safe assertions about conditionals

```
from pyon.util.arg_check import validate_is_true, validate_is_instance

validate_is_true(a, "A is not true.")

validate_is_instance(my_list,list,"My list is not a list.")
```


mmeisinger:
thanks for making it compliant. suggest merge when ready

PR #84, opened 2012-05-02
ION Process context flow of execution

daf:
- Container API refactored to allow multiple listeners to one IonProcess
- Listeners call into the service in a synchronized, safe fashion
- Error handling warped around to the correct calling greenlets
- Unit tests which cover the majority of usages

There will be some small cleanup to do afterwards - the ProcessSupervisor classes can be simplified a bit now that PythonProcess and friends are removed, but they are minor and unrelated to the main work here.

TBR @mmeisinger

### DIFF
--- a/pyon/container/cc.py
+++ b/pyon/container/cc.py
@@ -185,7 +185,7 @@ class Container(BaseContainerAgent):
         rsvc = ProcessRPCServer(node=self.node, from_name=self.name, service=self, process=self)
 
         # Start an ION process with the right kind of endpoint factory
-        proc = self.proc_manager.proc_sup.spawn((CFG.cc.proctype or 'green', None), listener=rsvc)
+        proc = self.proc_manager.proc_sup.spawn((CFG.cc.proctype or 'green', None), listeners=[rsvc])
         self.proc_manager.proc_sup.ensure_ready(proc)
         self._capabilities.append("CONTAINER_AGENT")
 
--- a/pyon/container/procs.py
+++ b/pyon/container/procs.py
@@ -167,8 +167,22 @@ class ProcManager(object):
         listen_name = get_safe(config, "process.listen_name") or service_instance.name
         log.debug("Service Process (%s) listen_name: %s", name, listen_name)
 
-        self._set_service_endpoint(service_instance, listen_name)
-        self._set_service_endpoint(service_instance, service_instance.id)
+        # Service RPC endpoint
+        rsvc1 = ProcessRPCServer(node=self.container.node,
+            from_name=listen_name,
+            service=service_instance,
+            process=service_instance)
+        # Named local RPC endpoint
+        rsvc2 = ProcessRPCServer(node=self.container.node,
+            from_name=service_instance.id,
+            service=service_instance,
+            process=service_instance)
+        # Start an ION process with the right kind of endpoint factory
+        proc = self.proc_sup.spawn((CFG.cc.proctype or 'green', None), listeners=[rsvc1, rsvc2], name=service_instance.id, proc_name=service_instance._proc_name)
+        self.proc_sup.ensure_ready(proc, "_spawn_service_process for %s" % ",".join((listen_name, service_instance.id)))
+
+        # map gproc to service_instance
+        self._spawned_proc_to_process[proc.proc] = service_instance
 
         # Directory registration
         self.container.directory.register_safe("/Services", listen_name, interface=service_instance.name)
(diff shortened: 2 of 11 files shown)


mmeisinger:
Quick skim pass done, more thorough will follow in a bit.
Looks very good. Thanks for cleaning up.

I'll have some follow up questions later. Here just some reminders for myself
- Can we separate process "thread" spawn from any endpoints? E.g. spawn_process (no endpoint) and then add_endpoint()?
- In procs, there still seem to be multiple spawns for the different endpoints. Intended? If yes (endpoint threads and process thread, I suggest renaming the operations to spawn_process, spawn_endpoint, add_endpoint or so)
- Make sure no higher level code calls proc_sub.spawn (this is only for ION processes)
- Process Supervisor should be for ION processes, not for greenlets. If the latter, then we should rename. Process is too overloaded and so is spawn. I would say ThreadSupervisor and start_thread or even something less thread-y, such as actor
- EventListeners should be the same as other endpoints.
- Only exception: incoming RPC can pass all (don't go into the process queue)
- Handling of greenlets as a framework function (in particular join and cleanup)
- Process input queue: how is it handled?
- One unack'd message at a time into the process queue

daf:
I'll try to answer the ones I understand.

> - Can we separate process "thread" spawn from any endpoints? E.g. spawn_process (no endpoint) and then add_endpoint()?

It's currently possible to do this via passing an empty list for the listeners kwarg, but having an `add_endpoint` would clean things up and allow adding listening endpoints after the fact (ie subscribers etc).

> - In procs, there still seem to be multiple spawns for the different endpoints. Intended? If yes (endpoint threads and process thread, I suggest renaming the operations to spawn_process, spawn_endpoint, add_endpoint or so)

This is intended - listening endpoints have their `listen` methods run in their own greenlets. Not sure what operations you want to rename here...

> - Process Supervisor should be for ION processes, not for greenlets. If the latter, then we should rename. Process is too overloaded and so is spawn. I would say ThreadSupervisor and start_thread or even something less thread-y, such as actor

`ProcessSupervisor` is indeed only for greenlets, and I agree we should rename.  The `pyon.container.procs.ProcManager` class is the manager/supervisor for ION processes.

Guessing that `ProcessSupervisor` was named due to original `PyonProcess` implementation.. it's a manager of those, not ION processes.

Would `GreenletSupervisor` be appropriate, or does that leak implementation?  `ThreadSupervisor` might work but is confusing.

> - EventListeners should be the same as other endpoints.

Adds weight to the `add_endpoint` change discussed above.

> - Only exception: incoming RPC can pass all (don't go into the process queue)

How would it achieve this bypass?  Header value in the message itself?

> - Handling of greenlets as a framework function (in particular join and cleanup)

I agree most people should not be using them directly, but need more explanation here.

> - Process input queue: how is it handled?

Listening Endpoints get a reference to an instance method in the `GreenIonProcess` (which should be renamed, there is no Green/Python distinction anymore).  They call this method which adds the real call to the synchronized queue in the ION process, and the control greenlet inside of every ION process will pick it up when available.

> - One unack'd message at a time into the process queue

This is not likely possible to be controlled.  Multiple listening endpoints will funnel a message into their handler whenever Pika notices that there is a message on the queue and there is not already an unacked message being processed.  This means that multiple listening endpoints attached to one ION process may get a single message from each of their queues and put it in the process queue to be called.

mmeisinger:
I like all of your comments. This is outside of the scope of this PR, but in the scope of the same task. Aggregated comments here

I would really like to split up:
- spawn ION process 
- spawn greenlet for an endpoint (or other purpose)
- add endpoint to process.

I would like to rename ProcessSupervisor to ThreadManager. It's still a threading programming metaphor that we apply, just not OS threads. Supervisor is reserved for ION process supervisor relationships.
So pyon.core.process becomes something else.
I don't like the name greenlet so much. Even gevent monkey patches the threading API, so that you can still use threading terminology.

I'd still like a separation of the pyon abstraction of threads (via ThreadManager) and the container use of threse, as a subclass.
Examples: IONProcessThread extends PyonThread, EndpointThread extends PyonThread, TimerThread extends PyonThread, BackgroundThread extends PyonThread

Add a ThreadManager.start() or start_thread() and converse for terminate/join/cleanup.

Good on the one thread per endpoint and one per process. 
Add a create_endpoint() to the container (or ExchangeManager) that calls ThreadManager.start() or add an endpoint.activate() that makes this call. I would really like to avoid higher layers calling a  ThreadManager.start() or the like.

Add a register_endpoint() to call to the container or process base class for the necessary wiring.

Definitely +1 to make EventSubscribers use endpoints and register them with the ION process.

Regarding threading/greenlet framework: ThreadManager will be the framework goto point.

Regarding the bypass: The RPC endpoint has its own thread. It does not place messages into the ION process queue, but directly notifies whatever AsyncEvent was waiting for the RPC response. The link-up is done by conversation-id and response queue. If a message comes into this specific response queue, the associated endpoint knows what to do.

Regarding one unack'ed message: this was per endpoint, only according to endpoint prefetch policy. The process queue can have as many messages in it as there are endpoints.

daf:
Following up tomorrow on action items via voice call.

shenrie:
Hi Dave,

I wanted to let you know that I am getting some pyon integration test
failures on my ubuntu installation that were not happening before I did a
pull a few minutes ago; however it is passing on my mac. I am attaching the
output file; I don't think it is a priority, however, it might be
representative of something more basic that is wrong.

Stephen

PR #86, opened 2012-05-04
`DotDict` Extension

lukecampbell:
Adds extension to `DotDict` 

Capabilities:
- Set attributes using Dot Notation

```
config = DotDict()
config.server_name = 'couch'
```
- Set attributes using chained Dot Notation

```
config = DotDict()
config.server.name = 'couch'
config.server.hostname = 'localhost'
config.server.port = '5984'
config.logging.verbose.failures = False
config.logging.verbose.exceptions = True
```
- Protection from overriding builtin methods

```
config = DotDict()
config.pop = 'Popped'
--> AttributeError: pop conflicts with builtin.
config.chained.attrs.pop = 'Popped'
--> AttributeError: pop conflicts with builtin.
```
- Unit Tests for new functionality

### DIFF
--- a/pyon/util/containers.py
+++ b/pyon/util/containers.py
@@ -56,18 +56,29 @@ class DotDict(DotNotationGetItem, dict):
     """
 
     def __dir__(self):
-        dictdir = dir(dict(self))       # woof, this is slow, but a rarely called method anyway, really for interactive work
-        dictdir.extend(self.iterkeys())
-        return dictdir
+        return self.__dict__.keys() + self.keys()
 
     def __getattr__(self, key):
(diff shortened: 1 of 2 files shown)


mmeisinger:
Looks good to me if thoroughly tested. Thanks for extending.

In general, I would advise to use DotDict as little as possible in favor if pure Python. It's more for the developer, providing simpler programming syntax than actually a data structure that does something useful and predictable.

It's internal complexity might lead to undetected bugs and it is less efficient than pure Python data type.

lukecampbell:
The applications that I see for this are with configurations, `CFG`, and mocking for unit tests.  In support of what you mentioned, I would not recommend this as a data structure.

PR #88, opened 2012-05-08
Added load_configuration.py to load configuration to the datastore

seman:


mmeisinger:
Thanks for this big refactor. Good progress. Here are my comments and requests:
- Can you place the code into a new pyon/core/interface/ package instead of pyon/util? I do want to keep pyon/util light and independent of any specific pyon concepts.
- Some hints regarding code conventions:
  -- classes always extend (object), the newer Python convention
  -- no space after a def methodname(params), never class Name:
- For the longer run, we should think of a better pattern for assembling files out of string tokens, instead of string concatenation.
- We should define some helper methods that are called from the generator functions for the standard patterns.

I recommend merge

PR #89, opened 2012-05-08
Actor refactor

thomaslennan:
UserIdentity usage in resource registry refactored to ActorIdentity.


mmeisinger:
See the ion-definitions PR comments. In general ok, but I don't agree with removing the UserInfo resource

PR #91, opened 2012-05-09
New granule

dstuebe:
Added New Granule for streams
Code is in: pyon/ion/granule
Example in: examples/granule.py

The granule is composed of a taxonomy and a record dictionary. There are helper classes: TaxyTool and RecordDictionaryTool which are used to operate on granule content. 

See: https://confluence.oceanobservatories.org/display/CIDev/R2+Construction+Data+Model
For design reference...

Update to COI Services pending with example Data Producer and Logging Subscriber.


mmeisinger:
Looks ok to me.

I wonder if in the longer run, this code (and the entire data model, science encoding stuff) should be in a separate repo that is added as submodule to coi-services. I don't think it really belongs in the container. Not every container will be concerned with data streaming and science data. I would like to see this plugged into the container in the longer run, but can live with it in pyon for now.

PR #92, opened 2012-05-11
Include xml files in eggs

oldpatricka:
Right now the production buildout config installs pyon as an egg (rather than with setup.py develop style deployment), and it doesn't include xml files in that egg. This causes problems with empty_policy_set.xml in one of the bootstrap levels in r2deploy.

This patch includes xml files in generated eggs.


daf:
Makes sense, thanks.

PR #94, opened 2012-05-11
Added code to handle decorators and inline comments

seman:


mmeisinger:
Can you add a unit test and an example to the PR please? I think we need to demonstrate correct behavior and exemplar use together with the code.

mmeisinger:
i suggest merge after checking that nothing currently breaks.

PR #96, opened 2012-05-14
New load test

dstuebe:
Added Test Runner for DM Stream load test using R2 (New) Granule


mmeisinger:
no issue.
as stated before, I would like to see the data model eventually move to a different repo

dstuebe:
I am happy to make the change. When do you suggest we split it out?

David

On Tue, May 15, 2012 at 5:23 PM, Michael Meisinger <
reply@reply.github.com

PR #99, opened 2012-05-18
Allow multiple connections to brokers

daf:
Connections to brokers are defined in `res/config/pyon.yml`, in two sections, `container.messaging.server` to get names like `primary` and `priviledged` [sic], and they key into connection information in `server`.

These changes move the concept of managing multiple nodes into the `ExchangeManager` class which every Container has.  Instead of `ExchangeManager.start` returning the default node and ioloop greenlet to be saved by the Container, it is now saved by the EM and indexable.
- If some connections fail but at least one succeeds, just warns about it.  If all fail or none are defined, throws an Exception
- `Container.node` is now a property that passes through to `ExchangeManager.default_node`, also a property, which returns `primary`, failing that any available, failing that a None.  Endpoints automatically get the `primary` still via this scheme.
- Exchange objects will use the `priviledged` [sic] connection if available.  (TODO: make sure EMS service uses it as well, it should)
- Full set of tests (unit/int)

TBR @mmeisinger, please don't merge until we've reviewed either offline or Monday.

### DIFF
--- a/pyon/container/cc.py
+++ b/pyon/container/cc.py
@@ -160,8 +160,6 @@ class Container(BaseContainerAgent):
                 os.kill(os.getpid(), signal.SIGTERM)
         self._normal_signal = signal.signal(signal.SIGTERM, handl)
 
-        self._capabilities.append("EXCHANGE_CONNECTION")
-
         self.datastore_manager.start()
         self._capabilities.append("DATASTORE_MANAGER")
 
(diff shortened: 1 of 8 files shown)


daf:
Forgot, this also adds two features @jamie-cyber1 requested/were on my long-term TODO:
- Delete pooled queues (rpc responses) when node is closed
- Delete anon queues from Subscribers on close (if it makes sense)

daf:
Merging after review with Michael + Tom over phone.

PR #101, opened 2012-05-24
CouchDB bulk association determination

lukecampbell:
Primarily intended to speed up and improve overall performance for association traversal in discovery.

`find_associations_mult` provides the capability to determine all the associations and objects belonging to a list of resources in a single query to CouchDB

```
><> org_id = '76f054f625804ee9a7198c2404c798fa'

><> org_manager = 'be31988aca3e495cbc9297741e2250d1'

><> pn.resource_registry.find_associations_mult([org_id])
--> 
[<interface.objects.UserRole at 0x103ca7810>,
 <interface.objects.ActorIdentity at 0x103ca7f90>,
 <interface.objects.ExchangeSpace at 0x103ca7b50>,
 <interface.objects.UserRole at 0x103ca7590>,
 <interface.objects.UserRole at 0x103ca7f10>]

><> pn.resource_registry.find_as
pn.resource_registry.find_associations       pn.resource_registry.find_associations_mult  

><> pn.resource_registry.find_associations_mult([org_id, org_manager])
--> 
[<interface.objects.UserRole at 0x103d270d0>,
 <interface.objects.ActorIdentity at 0x103d279d0>,
 <interface.objects.ExchangeSpace at 0x103ca7550>,
 <interface.objects.UserRole at 0x103d1de90>,
 <interface.objects.UserRole at 0x103d1d090>,
 <interface.objects.ActorIdentity at 0x103d1d1d0>]

```

### DIFF
--- a/pyon/datastore/couchdb/couchdb_datastore.py
+++ b/pyon/datastore/couchdb/couchdb_datastore.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python
-from pyon.util.arg_check import validate_is_instance
 
 __author__ = 'Thomas R. Lennan, Michael Meisinger'
 __license__ = 'Apache 2.0'
@@ -18,6 +17,7 @@ from pyon.datastore.datastore import DataStore
 from pyon.datastore.couchdb.couchdb_config import get_couchdb_views
 from pyon.ion.resource import CommonResourceLifeCycleSM
 from pyon.util.log import log
+from pyon.util.arg_check import validate_is_instance
 from pyon.core.bootstrap import CFG
 
 # Token for a most likely non-inclusive key range upper bound (end_key), for queries such as
--- a/pyon/datastore/couchdb/couchdb_datastore.py
+++ b/pyon/datastore/couchdb/couchdb_datastore.py
@@ -412,15 +412,14 @@ class CouchDB_DataStore(DataStore):
     def find_associations_mult(self, subjects, id_only=False):
         ds, datastore_name = self._get_datastore()
         validate_is_instance(subjects, list, 'subjects is not a list of resource_ids')
-        view_args = dict(keys=subjects)
-        if not id_only:
-            view_args['include_docs'] = True
+        view_args = dict(keys=subjects, include_docs=True)
         results = self.query_view(self._get_viewname("association","by_bulk"), view_args)
         ids = list([i['value'] for i in results])
+        assocs = list([i['doc'] for i in results])
         if id_only:
-            return ids
+            return ids, assocs
         else:
-            return self.read_mult(ids)
+            return self.read_mult(ids), assocs
 
 
 
--- a/pyon/datastore/couchdb/couchdb_datastore.py
+++ b/pyon/datastore/couchdb/couchdb_datastore.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python
+from pyon.util.arg_check import validate_is_instance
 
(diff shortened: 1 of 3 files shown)


mmeisinger:
Looks like a good addition. Thanks.
I encourage using X_mult operations where possible.
We just have to be careful that they keep the same guarantees than the individual operations do.

dstuebe:
This is a good optimization.

It raises an interesting point that I struggled with implementing associations in R1. You only get back a list of associated objects. There is no representation of what is associated to what in the result. In R1 I worked hard to make the in memory model represent the graph of associations as the basis for transaction that modify many resources and their associations. 

For a bulk query though - this is a great step.

mmeisinger:
Good point.

the result of this operation should be a 2-tuple of lists similar to the find_subjects etc operations [[], []]

The first list contains the object ids or objects, the second list contains the association objects that match. For one match the list index is the correlation

Is this possible?

lukecampbell:
yes

pyon/datastore/couchdb/couchdb_datastore.py
```diff
@@ -408,6 +409,21 @@ def _is_in_association(self, obj_id, datastore_name=""):
 
         return False
 
+    def find_associations_mult(self, subjects, id_only=False):
+        ds, datastore_name = self._get_datastore()
+        validate_is_instance(subjects, list, 'subjects is not a list of resource_ids')
+        view_args = dict(keys=subjects, include_docs=True)
+        results = self.query_view(self._get_viewname("association","by_bulk"), view_args)
+        ids = list([i['value'] for i in results])
```
mmeisinger, line 417: no need to write list( [ ] ). Just wrote [x for y in z], a list comprehension, one of the most powerful (and fast) features of python.

lukecampbell, line 417: oh, didn't know that, I thought iterables had to be type-casted to a list.

pyon/datastore/couchdb/couchdb_datastore.py
```diff
@@ -408,6 +409,21 @@ def _is_in_association(self, obj_id, datastore_name=""):
 
         return False
 
+    def find_associations_mult(self, subjects, id_only=False):
+        ds, datastore_name = self._get_datastore()
+        validate_is_instance(subjects, list, 'subjects is not a list of resource_ids')
+        view_args = dict(keys=subjects, include_docs=True)
+        results = self.query_view(self._get_viewname("association","by_bulk"), view_args)
```
mmeisinger, line 416: there is another function find_by_view() that should be used if possible for any view operations. It's mostly redundant in intent to the query_view/parse_results and the expected way of doing things.

1: Feel free to carefully extend find_by_view with missing capabilities (without breaking things)

2: This is a potential connection point to custom indexes for objects (we need to discuss how RR and DM index mgmt can be integrated)

pyon/datastore/couchdb/couchdb_datastore.py
mmeisinger, line 417: I suggest reading about list comprehensions and generator expressions. If you want to boost your Python level to expert and gain a order of magnitude of performance, this is the way to go (no kidding).

Btw, you did it, just copied one list into another

[x for y in z]  returns a list
(x for y in z) returns a generator

so you would have been right for list( (x for y in z) ) - note dual parens - but a simple list comprehension is better.

Some people try to write code completely in this style. Thats a bit over the top

PR #102, opened 2012-05-31
Check to ensure service deps exist when starting other services

daf:


mmeisinger:
Thanks, that's all there is to it (initially). Good, simple impl. Just completing our checks to detect errors early.

There will be other tasks complementing this, but outside of this task.

Remember to log time against this one (be generous)

PR #106, opened 2012-06-07
Interceptor Redo

daf:
Removes module initialization for interceptors, cleans a variety of things up.

Interceptors belong to the Node, because Endpoints can be used outside of a Container.  They are initialized on startup (Exchange Manager start), and propagate to all Endpoint/EndpointUnit.

Endpoints expose the interceptors through properties on themselves, where they can either be set on a per endpoint/endpoint unit level, or just take the system default. This is very flexible in operation.

### DIFF
--- a/pyon/core/bootstrap.py
+++ b/pyon/core/bootstrap.py
@@ -101,10 +101,6 @@ def bootstrap_pyon():
     from pyon.ion import resource
     resource.load_definitions()
 
-    # Load interceptors
-    from pyon.net.endpoint import instantiate_interceptors
-    instantiate_interceptors(CFG.interceptor)
-
     # Services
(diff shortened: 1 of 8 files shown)


shenrie:
I can say for AS I think we need at a minimum the following to display
visualizations:
- Dave Foster to have the exchange objects which represent queues per our
  discussion this week (for historical realtime plots)
- Ingestion and "replay" from DM, both for the graduals for the science
  data  ( for historical plotting/graphs) and the "blob" graduals ( which
  would represent the matplotlib generated images)

Michael, is this in agreement with what we discussed yesterday with Don?

Stephen

---

System Architect

daf:
Email discussions seem to think this is a good idea. Merging.

PR #109, opened 2012-06-15
Fix getting password for couchdb_standalone.

oldpatricka:
Looks like a copy/paste error. Fixes the last of my problems setting the couchdb connection details with the --config option.


mmeisinger:
sorry

PR #113, opened 2012-06-27
adding DB repository layers

newbrough:
initial classes for splitting couchdb_datastore into IonObject-specific repository.py and dict-only db-specific couch_store.py.  also includes connection pooling and shorter and sequential ID generation.


mmeisinger:
This is a good start. Merge for now and let's sync how to bring this in use

PR #115, opened 2012-06-29
Luke: adds the ability to combine granules

lukecampbell:
- Primitive combination of granules
- Moves the granule classes into the init file for easier imports

You may now import the granule wrappers from `pyon.ion.granule` directly:

```
from pyon.ion.granule import RecordDictionaryTool, TaxyTool, build_granule, combine_granules
```

### DIFF
--- a/pyon/ion/granule/taxonomy.py
+++ b/pyon/ion/granule/taxonomy.py
@@ -260,4 +260,7 @@ class TaxyTool(object):
 
         return False
 
+    def __ne__(self, other):
+        return not self == other
+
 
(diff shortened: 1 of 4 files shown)


pyon/ion/granule/taxonomy.py
```diff
@@ -260,4 +260,7 @@ def __eq__(self, other):
 
         return False
 
+    def __ne__(self, other):
+        return not self == other
+
```
lukecampbell, line 265: if you implement `__eq__` you should also implement `__ne__` as per [Python DataModel documentation](http://docs.python.org/reference/datamodel.html)

PR #126, opened 2012-07-16
Luke: adds simple stream subscriber and publisher

lukecampbell:


daf:
This is the first step in a refactor of the existing StreamPub/Sub objects - currently they are too specific.

PR #127, opened 2012-07-20
Luke: fixes shape bug with rdt

lukecampbell:


lukecampbell:
Shouldn't affect anyone it's only additive.  It will kill any code that violates an array shape with granules though, which is good because we can't have mis-shaped granules!

PR #140, opened 2012-08-08
Luke: adds change_logging_level

lukecampbell:
@daf 
@mmeisinger
@mauricemanning
@jamie-cyber1

This adds the ability to interactively change the logging level, it's intended use is for debugging interactively.
Example:

```
from pyon.util.log import change_logging_level
change_logging_level('pyon','CRITICAL')
```

### DIFF
--- a/pyon/util/log.py
+++ b/pyon/util/log.py
@@ -12,6 +12,16 @@ import sys
 import socket
 from logging.handlers import SysLogHandler, SYSLOG_UDP_PORT
 
+
+DEBUG     = 'DEBUG'
+INFO      = 'INFO'
+WARNING   = 'WARNING'
+ERROR     = 'ERROR'
+CRITICAL  = 'CRITICAL'
+EXCEPTION = 'EXCEPTION'
+
+
+
 class StackFormatter(logging.Formatter):
     def __init__(self,*a,**b):
         super(StackFormatter,self).__init__(*a,**b)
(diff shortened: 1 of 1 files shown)


daf:
Hold off for now, Luke's adding the ability for individual loggers.

newbrough:
Don't we already have this with log.setLevel ?

lukecampbell:
Where?

```

(work)Lukes-ASA-Macbook:pyon (time %) luke$ ack 'setLevel' pyon
pyon/util/test/test_exception_logging.py
21:        self.subject.setLevel(INFO)
```

newbrough:
Where what?

Maybe my last email wasn't clear...

The pull request comment was "hold off, Luke is adding the ability to
change [logging level] of individual loggers"

And my response to that was, can't we already do that using something like
this:

from pyon.util import log
log.setLevel(INFO)

Jonathan

lukecampbell:
It doesn't change the logging level of an individual logger.

newbrough:
which "it"?  my example or your code change?

lukecampbell:
This code allows for an operator to interactively change the logging level for a specific logger. The method takes two parameters the desired logger and the logging level. It is intended for interactive debugging purposes. 

For example: If you are working on coi-services and want to enable debug information on your ion module in coi-services you could run the following:

```
from pyon.util.log import change_logging_level, DEBUG
change_logging_level('ion',DEBUG)
```

This changes the logging level for only the _ion_ logger while leaving the other loggers unchanged.

newbrough:
import logging
logging.getLogger('ion').setLevel(logging.DEBUG)

newbrough:
would be nice to see something that re-reads the logging.yml files.  i
could limit logging until i got to the critical thing i was testing, then
have use reset_logging('logging.alternate.yml') to change lots of levels at
once.

lukecampbell:
I followed PEP8
:P

newbrough:
which part?

lukecampbell:
method name: https://github.com/ooici/pyon/pull/140/files#L0R221

The other consequence of this method is that the internal global `LOGGING_CFG` is properly updated to reflect the change, although I don't know if anything else uses it. Honestly, I had no idea you could change it with logging which is what you're probably dying to hear considering the nature of this belabored commentary on a closed pull request. 

In any case it's in.

PR #148, opened 2012-08-31
Luke: adds stream route pub/sub

lukecampbell:


pyon/ion/stream.py
```diff
@@ -204,6 +207,24 @@ def new_publisher(cls,container,exchange_point,stream_id):
...
+    def __init__(self,exchange_point, routing_key):
+        validate_is_instance(exchange_point,ExchangePoint)
+        self.routing_key = routing_key
+        self.exchange_point = exchange_point
+        super(SimpleStreamRoutePublisher,self).__init__()
+
+    def publish(self, msg, routing_key=''):
+        if not routing_key:
+            routing_key = self.routing_key
+        return super(SimpleStreamRoutePublisher,self).publish(msg,to_name=self.exchange_point.create_route('%s' % routing_key))
```
daf, line 220: Why bother with the formatting syntax?  Is `routing_key` not going to be a string already?  If not, why not just `str()` it instead?

pyon/ion/stream.py
```diff
@@ -225,3 +246,25 @@ def new_subscriber(cls, container, exchange_name, callback):
...
+        self.close()
+        self.greenlet.join(timeout=10)
+        self.started = False
+    @classmethod
+    def new_subscriber(cls, container, exchange_name,stream_route,callback):
+        xp = container.ex_manager.create_xp(stream_route.exchange_point)
+        xn = container.ex_manager.create_xn_queue(exchange_name)
+        xn.bind(stream_route.routing_key, xp)
+        instance = cls(name=xn,callback=callback)
+        setattr(instance,'xn',xn)
```
daf: Why `setattr` instead of `instance.xn = xn`?

pyon/ion/stream.py
lukecampbell, line 220: Mostly a "just in case" kind of habit. Such that if a client passes in an object which has a string form. like uuid or something.

PR #152, opened 2012-09-20
Luke: adds numpy capabilities to msg encoder

lukecampbell:


shenrie:
Will adding numpy capabilities to the message encoder require that numpy is still a dependency in pyon? Or can it be done in a safe way that if numpy is not installed and not ever used, then the encoder will still work without it?

lukecampbell:
Numpy already was a dependency because of the encoder's previous capabilities.

shenrie:
I understand that, but I think we were hoping there would be a way to remove this dependency on numpy in pyon if possible ( similar to HDF5).

daf:
I think what Stephen is alluding to is the ability to have numpy be an OPTIONAL dependency for pyon.  I see these steps:
- user builds out with a different buildout.cfg for a certain environment (`buildout_no_numpy.cfg` or `buildout_gumstix.cfg` etc)
- the code in this file reacts appropriately to numpy being present or not present.  an inline import is a possibility (slow if the lib is not present as it would likely take that path every time), could likely come up with a better solution.  maybe a separate interceptor, derived from this interceptor, which has the numpy related code in there.  then it's the config files responsibility to select the proper interceptor to use.

Am I characterizing this correctly?

lukecampbell:
If that's the case then I can make some try/excepts around the numpy usage in Pyon to assist in this goal.

blazetopher:
The current plan as I understand it, is that the coverage model and persistence layer (i.e. numpy and HDF) will both be present on the gumstix.  If that is the case - is there still a case where we would want to build without numpy/hdf?

shenrie:
Yes I think there is a case for it, because its not just about the gumstix. We anticipate there being other uses for light weight containers that won't necessarily need to encode/decode science data. 

I like the idea of dynamically invoking interceptors as needed over try/excepts...I think that is more inline with Micheal's architectural view.   Another thought along these lines, is why do all of the interceptors have to be located in pyon? I think it should be possible to have coi-services define interceptors that may be used for conveying science data but could have different technologies, like numpy, vs those in the base pyon instance.

daf:
Agree fully with the split.  Currently it's not really possible but that's more of just tweaking configuration to allow appending interceptors etc - the framework is there, just not how to configure it.

Having numpy support all the way down at the pyon level does not feel correct, as nothing else in pyon really needs it (from what I can think of currently).

And yes, I was also thinking of non-gumstix scenarios where we would want a stripped and reduced build.

shenrie:
It could be just a difference in using different pyon.yaml config files as well; ones that contain a different set of interceptors. There should be no expectation that a lightwight install, vs a gumstix install vs an install on an ION node should use the same config file.  I just don't think we have accounted for that in our dev environment since the ion-defs is shared across both projects and it may be time to diverge the config files for pyon vs coi-services.

lukecampbell:
[ion_time](https://github.com/ooici/pyon/blob/master/pyon/util/ion_time.py) uses it... I think we can probably get away with moving this somewhere else though.  It probably belongs in ion/core/util or somewhere like that.

daf:
As we've already got a numpy dep in pyon right now, let's merge this to move forward, and make a note to clean it up as part of a numpy sweep, as part of https://jira.oceanobservatories.org/tasks/browse/CIDEVCOI-610 probably.

shenrie:
Agreed. I think that is the task Seman is working on

lukecampbell:
I'm working with Seman as we speak to refactor out some deprecated stuff to get rid of h5py and numpy.

PR #155, opened 2012-09-21
pyon/ion code cleanup

seman:


mmeisinger:
The changes look good towards PEP8 compliance. Please wait to merge for a discussion with Tim, to see if this is a good time to do it right now or later.

daf:
I MUCH prefer aligning lists of assignments etc on `=` or `:` characters; this removes that in several places.  Code is much more readable with them in.  Does it violate PEP8 or some other standard?

lukecampbell:
I don't believe it does, the only PEP8 convention which may be violated is the 80 character limit per line but we have to violate it in several places because our project is deep and our service implementations have long names.

mmeisinger:
I will talk with Tim and Jamie about it. I'm against following a convention such as PEP8 to the last word, if there are good reasons, such as legibilty or clarity of intent.

daf:
We already can't/don't, due to the auto-gevent-patch from importing anything pyon.  This means pyon imports should be above stock imports.

mmeisinger:
And we do not intent to enforce the 80 character rule, see https://confluence.oceanobservatories.org/display/CIDev/Code+Conventions

lukecampbell:
and I would like to point out the built in `unittest` module as well as others violates PEP8 with the naming conventions. And we follow that.

seman:
fyi, I am using http://pypi.python.org/pypi/pep8 integrated with PyCharm as
external tool to check for code convention.

mmeisinger:
Discussed. This PR will be closed and Seman will prepare a new PR with less extensive PEP8 changes (see comments)

PR #158, opened 2012-09-27
Luke: Makes utilities a submodule

lukecampbell:


lukecampbell:
Unbeknownst to me we're not going down this road...

PR #163, opened 2012-10-17
Adds configurable durability to Exchange Object based queues

daf:
For the running system, we want the ability to have durable queues for system failures.  This branch adds this as a configuration setting, `container.exchange.names.durable: True`.

This config setting only affects queues created via the Exchange Object API, which is used by Streams, Pubsub, and other high level, data communication pieces of the system.  It does _NOT_ affect service or process queues, intentionally.
#### Some highlights of the PR:
- Able to set durability via configuration.  Currently in ion-defs it is `False`, but you may test by placing this block in your `pyon.local.yml`:

``` yaml
container:
  exchange:
    names:
      durable: True
```
- Adds ability to `--broker_clean` on the pycc command line.

The `--broker-clean` or `-bc` flag will remove any exchange or queue starting with the same sysname that this container is starting up with.  It is functionally similar to the `--force_clean` option that exists for CouchDB.  Combine with `-X` for immediate exit, you can use the pycc entry script as an ad-hoc cleaning utility.

Example use:

```
~/dev/code/pyon -> ./bin/pycc --rel=res/deploy/examples/hello.yml system.name=wizzzard -bc -X
pycc: ION Container starter with command line options: Namespace(clean_broker=True, config=[], config_from_directory=False, daemon=False, force_clean=False, immediate=False, logcfg=None, mx=False, no_container=True, noshell=False, pidfile=None, proc=None, rel='res/deploy/examples/hello.yml', signalparent=False, sysname=None)
pyon: sys_name changed from 'None' to 'ion_ASA-DEVELOPERs-MacBook-Pro_local'
pyon: sys_name changed from 'ion_ASA-DEVELOPERs-MacBook-Pro_local' to 'wizzzard'
pycc: auto_bootstrap=True.
pyon: pyon.bootstrap (bootstrap_pyon) executing...
pycc: clean_broker=True, sysname: wizzzard
      exchanges deleted (2): wizzzard,wizzzard.pyon.events
         queues deleted (1): wizzzard.hello
pycc: no_container=True. Stopping here.
```
- Adds properties for `queue_durable`, `queue_auto_delete`, `exchange_durable`, `exchange_auto_delete` from channel layer. Now correctly reports this value if an exchange object is used for a channels send or recv name.
- Sets proper `delivery_mode` flag for message retention by broker in durable queues.  Method for discovering if it should be sent with this mode could use improvement, but it's due to the intentional disconnect between senders and receivers in rabbit, and our heterogenous use of durable/non-durable queues.
#### Less important highlights:
- Adds some new Rabbit Exchange Manager HTTP API functionality: `delete_queue` and the ability to pass in data to be included as part of the request, only useful if manually calling `_make_management_call` with `post` or similar
- Extracts `RabbitManagementHelper` and utility methods from the pycc plugin and moves them into putil for use by `scripts/pycc`


daf:
This PR intended for code review.  Please do not merge it without discussion/looking at code.

@mmeisinger, @shenrie, @jamie-cyber1

lukecampbell:
Any `StreamProcess` which is launched has a non-service queue for processing incoming data off of a stream. These queues are created using exchange objects, therefore they are set to durable (as they should) when other XN queues are durable. A very nice consequence of this PR. 

https://github.com/lukecampbell/pyon/blob/22423a244101037d39eb1cf2ee072c4969e84cac/pyon/container/procs.py#L414-414

daf:
SMOKE tests pass locally with this on.

PR #170, opened 2012-11-02
Integrating the Conversations API and Interceptor

shenrie:
This pull request should include all of the pyon changes needed to integrate the Conversations API and Interceptor which runs on top of the existing ION process endpoint code. Once this is merged, then Conversations will be automatically enabled for all RPC requests because of a pyon.yaml configuration of rpc_conversation_enabled set to True by default and by generating the service interface files. 

To disable Conversations, set rpc_conversation_enabled to False and re-generating the service interface files. 

UNIT and coi-services SMOKE tests all passed but I would still like to request a review of all the changes, especially to the existing container and endpoint code before fully merging. Please do not merge without letting me know when you are going to do it.


pyon/container/cc.py
```diff
@@ -199,7 +199,7 @@ def handl(signum, frame):
         self._is_started    = True
         self._status        = "RUNNING"
 
-        log.info("Container started, OK.")
+        log.info("Container (%s) started, OK." % self.id)
```
daf: Not important here, but use the comma form of log instead of %.

```
log.info("Container (%s) started, OK", self.id)
```

prevents unneeded string interpolation when log levels aren't set.

pyon/container/procs.py
```diff
@@ -347,7 +348,11 @@ def _create_listening_endpoint(self, **kwargs):
             eptype          = getattr(mod, cls)
             ep              = eptype(**kwargs)
         else:
-            ep = ProcessRPCServer(**kwargs)
```
daf, line 350: Looks like we're using this block instead of the configurability block I put in - should remove it from ion-defs.

shenrie, line 350: I wanted to leave that there so we have some flexibility if we need to change it. I think its harmless to leave in for now.

pyon/net/endpoint.py
```diff
@@ -826,7 +836,8 @@ def _build_header(self, raw_msg, raw_headers):
         """
         headers = BidirectionalListeningEndpointUnit._build_header(self, raw_msg, raw_headers)
         headers['performative'] = 'inform-result'                       # overriden by response pattern, feels wrong
-        if self.channel and self.channel._send_name and isinstance(self.channel._send_name, NameTrio):
+        #TODO - figure out why _send_name would not be there
+        if self.channel and hasattr(self.channel, '_send_name') and self.channel._send_name and isinstance(self.channel._send_name, NameTrio):
```
daf, line 840: Will the conversations codepath know how to reply if this header is not set?  It's required for normal endpoint operations but conversations probably supercedes it.

daf:
Reviewed main integration points with existing endpoint/messaging/container code.  Only small comments, nothing to address.  I would like to do a more thorough look at this next week if possible.

pyon/net/endpoint.py
shenrie, line 840: This might have been a remnant of using Rumi's old channel stuff, but now that I am using the normal endpoints, it should always be there. I was just adding the hasattr(self.channel, '_send_name') to be safe"

pyon/ion/conversation.py
```diff
@@ -0,0 +1,406 @@
...
+        return self._endpoint._participant
+
+    @property
+    def conv_type(self):
+        return self._endpoint._conv_type
+
+    def _message_send(self, msg, headers=None, **kwargs):
+        return ProcessRPCRequestEndpointUnit.send(self, msg, headers,  **kwargs)
+
+    #TODO - discuss with Dave to make sure this is the best way to hook into the returned message
```
daf: This is good :thumbsup:

pyon/container/cc.py
shenrie: Not sure I understand how the comma is different that the %

pyon/ion/conversation.py
```diff
@@ -0,0 +1,406 @@
...
+
+    @property
+    def participant(self):
+        return self._endpoint._participant
+
+    @property
+    def conv_type(self):
+        return self._endpoint._conv_type
+
+    def _message_send(self, msg, headers=None, **kwargs):
```
daf, line 289: This looks a vestigial remnant of something else, consider removing just in case this method signature changes or something.  Also the kwarg to positional arg translation could be an issue.

pyon/ion/conversation.py
```diff
@@ -0,0 +1,406 @@
...
+
+        if hasattr(self._process, 'name'):
+            self._participant = Participant(self._process.name)
+        else:
+            self._participant = Participant(self._send_name)
+
+
+
+    def create_endpoint(self, to_name=None, existing_channel=None, **kwargs):
+        return ProcessRPCClient.create_endpoint(self, to_name, existing_channel, **kwargs)
```
daf, line 336: Same comment as above with a remnant.

pyon/container/procs.py
shenrie, line 350: I don't see where it is in the ion-defs cause I think we all added it to pyon.local.yaml

pyon/ion/conversation.py
shenrie, line 289: These are short cuts to get at the Participant and ConversationType objects which are created by the endpoints; not the endpoint units so not remnants at all.

pyon/ion/conversation.py
shenrie, line 336: Same reply as above :)

PR #186, opened 2012-11-16
Use the Service status field to determine service availability

shenrie:
This utilizes the new status field added by the CEI HA Agent to show the status of the service based on the availability of a worker process. This is intended to fix the timeouts seen in the integration tests that use multiple containers that occur when the policy service is not fully available.

### DIFF
--- a/pyon/ion/exchange.py
+++ b/pyon/ion/exchange.py
@@ -255,7 +255,7 @@ class ExchangeManager(object):
             if exchange_service:
                 return True
 
-            return False
+        return False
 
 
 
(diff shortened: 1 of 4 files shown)


pyon/ion/exchange.py
```diff
@@ -250,11 +251,13 @@ def _ems_available(self):
...
         if CFG.get_safe('container.exchange.auto_register', False) and self.use_ems:
             # ok now make sure it's in the directory
-            service_list, _ = self._rr.find_resources(restype="Service", name='exchange_management')
-            if service_list is not None and len(service_list) > 0:
+            exchange_service = get_service_registry().is_service_available('exchange_management')
+            if exchange_service:
                 return True
 
-        return False
+            return False
```
daf: This must be moved back an indent level - needs to be outside of the if block.

shenrie: Good find....fat finger ;)

PR #188, opened 2012-11-20
Transpose ResourceAgentClient name to to_name

daf:
Fix straggler to update to `to_name`, but `name` looks rather baked into `ResourceAgentClient`


mmeisinger:
no issue looks good

PR #189, opened 2012-11-27
resource_registry property in service clients to use container if available

daf:
Indirects out `self.clients.resource_registry` to be a property that checks to see if the container has a resource registry client available and uses that.  If not available, uses its own process client.

There is one missing method available in the `ResourceRegistryService`: `get_resource_extension` - it's not used by any services currently in coi-services, but these interfaces should be aligned to prevent any future problems.


mmeisinger:
looks good

daf:
Will bump coi-services along with factoring out the `has_capability` pattern we introduced a few weeks ago to use this new indirection.

PR #194, opened 2012-12-05
removes special case of read_mult on [ ] by simply returning [ ]

ianfixes:
This is NOT UNIT TESTED but works in practice.

This commit removes a special case in read_mult -- for any number of resource_ids, read_mult returns that number of resources.  Previous behavior was to raise an error if 0 resource_ids were supplied.  The previous behavior seems sensible, unless you consider cases where a search algorithm might gather some list of resource_ids then want to convert them all to resources.

The new behavior makes option A equivalent to option B, which is what you'd expect:

``` python

resource_ids = find_some_resource_ids()

#option A

resource_objs = resource_registry_client.read_mult(resource_ids)

#option B

resource_objs = [resource_registry_client.read(r) for r in resource_ids]

```

### DIFF
--- a/pyon/datastore/couchdb/couchdb_datastore.py
+++ b/pyon/datastore/couchdb/couchdb_datastore.py
@@ -296,6 +296,7 @@ class CouchDB_DataStore(DataStore):
         return obj_list
 
     def read_doc_mult(self, object_ids, datastore_name=""):
+        if not object_ids: return []
         ds, datastore_name = self._get_datastore(datastore_name)
         log.debug('Reading head version of objects %s/%s' % (datastore_name, object_ids))
         docs = ds.view("_all_docs", keys=object_ids, include_docs=True)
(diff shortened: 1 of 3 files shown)


daf:
I have to believe this was the original intention of the code - just happened to be caught by the falsiness of `[]` instead of explicitly checking for None as you've changed.

That being said, this needs to be tested and tested HARD.  This is a very common operation, so we'd probably need to run the full gauntlet of both INT and UNIT tests for both pyon and coi-services.  `--with-pycc` would be fine to help speed it up.

ianfixes:
Well, if you only modify the check for `None` vs empty list (without the change I made to couchdb_datastore.py where I check for `[]`), it produces a couch error.  So presumably there is a reason not to give an empty list to couch, but read_mult shouldn't have that restriction.

daf:
I agree, no reason to go to couch when you know it's empty anyway.

ianfixes:
OK.  So just to be clear, it was not the designers' intention for `read_mult([])` to raise an exception, and this should be fixed.  I can run that test in a bit.

shenrie:
You should run this by Michael as I think he originally implemented this code and has some strong thoughts about how the RR functions should work and what they return/throw.

mmeisinger:
I think the proposed change is OK. I've been annoyed by this behavior many times myself.

read_mult([]) should return [] not exception.

Of course, it should not go the the database with an empty query set.

PR #197, opened 2012-12-10
Attaches stream publishing endpoints before on_start is called

lukecampbell:
Part of [OOIION-594](https://jira.oceanobservatories.org/tasks/browse/OOIION-594)


daf:
Looks good.

PR #200, opened 2012-12-11
Fix OOIION-602: make RR wrapper to preserve ion-actor-id as owner

mmeisinger:


mmeisinger:
Dave, FYI, to be enhanced if needed

PR #202, opened 2012-12-11
Call close on Publishers attached via proc manager

daf:
- moves around location of calls to attach publisher
- makes "simple" process have an IonProcessThread, mostly for
  cleanup purposes

Fixes OOIION-595


mmeisinger:
Looks good

PR #206, opened 2012-12-14
Fixes StandaloneStreamPublisher

lukecampbell:
StandaloneStreamPublishers were making an XP during every publish call, this addresses it.


daf:
:ok_hand:

PR #210, opened 2012-12-24
Fix OOIION-625: Scalable extended resources queries

mmeisinger:
This is an updated version of the PR with fixes that broke tests


mmeisinger:
Ok, it's time, I've skipped the unit test (Stephen may help adjusting it to the new impl), checked the EXT int tests and checked it via the UI.

PR #211, opened 2013-01-02
Reduces memory consumption in stream pubsubs.

lukecampbell:
The XP route objects were being unneccesarily created for every publish
action, the object is now created during instantiation and simply
referenced during the lifespan of the publisher. This eliminated some
leakiness.


daf:
:+1:

PR #216, opened 2013-01-09
Updated generate interface for a dry run option 

seman:
Updated message_object_generator.py and object_model_generator.py so that during a dry run, it won't write to any files


mmeisinger:
looks good

PR #224, opened 2013-01-16
Stops creation of folders during cleaning only operations.

lukecampbell:


daf:
:cactus:

lukecampbell:
I'm guessing :cactus: means we're :cool:.

PR #226, opened 2013-01-23
Add auto_delete to EventSubscriber initializer (OOIION-676)

daf:
Adds the ability to pass auto_delete up to the channel creation for a listening `EventSubscriber`/`ProcessEventSubscriber`.

In the case of a clean shutdown, this is redundant: the `SubscriberChannel` close already will delete any anonymous queues (given it meets a long list of conditions), but the auto_delete flag here will handle non-graceful shutdowns.  See https://github.com/ooici/pyon/blob/master/pyon/net/channel.py#L954-L957 for implementation details on the Subscriber's deletion mechanism.


labisso:
This looks like what we need for HA Agents. I didn't realize that cleanup was already happening for a clean shutdown, but that makes sense. Our orphaned queues resulted from total container failures.

PR #231, opened 2013-01-24
Exception event

mbarry02:


pyon/event/event.py
```diff
@@ -323,3 +325,18 @@ def await(self, timeout=None):
...
+        fn()
+    except StreamException as e:
+        info = "".join(traceback.format_tb(sys.exc_info()[2]))
+        pub = EventPublisher(event_type="ExceptionEvent")        
+        pub.publish_event(origin="stream_exception", description="stream exception event", exception_type=str(type(e)), message=info)
+
+def handle_stream_exception(fn):
+    def wrapped():
+        _handle_stream_exception(fn)
+    return wrapped
```
lukecampbell, line 344: The function `fn` needs to be able to handle any number of arguments, i.e. `(*args, **kwargs)` for [example](https://github.com/lukecampbell/pyon/blob/0fb740bdfc73688fd0ac67e92e05137ee90d9d37/pyon/util/memoize.py#L30-59)

```
def handle_stream_exception():
  @functools.wraps(fn)
  def wrapper(*args, **kwargs):
    ...
    fn(*args, **kwargs)
    ...
  return wrapper
```

pyon/event/test/test_event.py
```diff
@@ -239,6 +243,47 @@ def cb(*args, **kwargs):
...
+            if self.count == 3:
+                ar.set()
+
+        #test file system error event
+        sub = EventSubscriber(event_type="ExceptionEvent", callback=cb, origin="stream_exception")
+        self._listen(sub)
+
+        def _raise_filesystem_error():
+            raise FilesystemError()
+        _handle_stream_exception(_raise_filesystem_error)
```
lukecampbell: Can you try in your test to declare it with the wrapper like:

```
@handle_stream_exception
def _raise_filesystem_error():
  raise FileSystemError()
```

That way the actual decorator is also tested.

lukecampbell:
Reviewed, waiting on a few minor changes.

lukecampbell:
Merged by hand, github is slow.  No Conflicts

PR #234, opened 2013-01-25
Health check for bidir pooled channels (OOIION-693)

daf:
See bug for more description: https://jira.oceanobservatories.org/tasks/browse/OOIION-693

Relevant pika bug: https://github.com/pika/pika/issues/263

Issues:
- this _WILL NOT_ solve the same situation with EventSubscriber that's been observed
- the queue will not be cleaned up on container termination - although that would happen on abnormal container termination etc, need separate sweeper for broker


shenrie:
Is there a concern that the dead_pool could grow too large? Seems like the container should not continue to operate if there are so many dead channels accumulating.

daf:
Honestly, hard to tell the frequency of this problem. I put the `dead_pool` item in there so we can manhole our way in when we notice the WARN message and poke at things.  I intend to remove it after we get a little more clarity as to what is going on.

PR #243, opened 2013-02-07
Refactored event publishing APIs

shenrie:


mmeisinger:
Excellent. Looks good.

A few comments on pyon/event/event.py:

Line 66: Python does not allow 2 functions with the same name. The lower overrides the above. Just delete.
Line 132: I would make the check go both ways. Too old or too new timestamp is both bad.
Line 133: Use log.error instead. Log.exception expects that you are in an except: block
Line 143: I would forbid an _id already being set by the caller. Also, this line is a bit redundant
Line 180: Better to return the event object instead of a bool.

pyon/util/containers.py is_valid_ts:

Please also check:
- Type of original argument is str (or better basestring)
- String value has no whitespace. E.g. int("  123213  ") is is not an exception but an invalid value.
- Please also check that it is exactly 13 chars long, and non negative. This rules out mistakes of using seconds instead of millis

shenrie:
I intended to delete these other functions, just left them around until everything was finalized.

The log.exception() is just a copy/paste error, so good catch.

I wasn't sure if there might ever be a situation where an _id would exist already, but if you don't think that should ever be allowed then even better.

The return values are what you had previously defined; I was not planning on changing this and not sure if any callers are depending on this already.

shenrie:
Looks like basestring is not supported in Python 3.x...not sure if that is a concern or not.

lukecampbell:
Not sure if it matters or if we should go down this path just yet, but python/python3 string checking can be done as:

``` python
try:
    return isinstance(obj, basestring)
except NameError:
    return isinstance(obj, str)
```

shenrie:
Since I am in here...does anyone know why we are doing this?

def current_time_millis():
    return int(time.time() \* 1000)

versus this:

def current_time_millis():
    return int(round(time.time() \* 1000))

The latter seems to be the accepted recommendation on the net...

mmeisinger:
No major reason when I implemented this. A small reason was the visual prefix nature of timestamps:

I thought it would be better to display:
123.123567 s -> 123.123

than the more precise, but also more confusing
123.123567 s -> 123.124

You basically truncate a higher precision string by cutting away digits.

shenrie:
yes, but isn't that what the round() is for? to ensure a more accurate value?

shenrie:
And should the timestamp really only be checked to be a length of 13 digits and not 13 or greater? At some point perhaps not in our lifetime, the timestamp will become 14 digits; or will we have been annihilated by nuclear war or an asteroid before that happens?

mmeisinger:
Correct. I did not think of it back then (or thought it not important for the early prototyping).

But yes, I'm not terribly content with these ION timesteps - just don't want to change them now.

Note that we only use ION timestamps in the pyon infrastucture, not for science measurements where we use NTP4. Millisecond accuracy seemed fine at this time and I believe still is.

We could round now. I don't see any implications to your suggestion.

mmeisinger:
UNIX timestamps will flip to decimal 14 digits in the year 2286

> <> time.localtime(9999999999)
> --> time.struct_time(tm_year=2286, tm_mon=11, tm_mday=20, tm_hour=9, tm_min=46, tm_sec=39, tm_wday=5, tm_yday=324, tm_isdst=0)

Btw they flipped to 13 digits in the year 2001. 

This time period if 13 character times in mlliis seems large enough for now and we can rule out a class of errors bu having better checks.

shenrie:
So the answer is 'B'...we will have been annihilated by nuclear war or an asteroid before that happens :)

mmeisinger:
Let's make it 13 or 14 digits then permissible length. This will postpone the nuclear annihilation to year 5138.
Our ancestors can then build a new pyon util layer. :)

mmeisinger:
Excellent. I suggest merge when ready

shenrie:
Did you see my email about exceptions vs log statements?

shenrie:
I agree that the return type of the API should be the event object which is created( since it would enable the caller to refer to the _id or ts_created for completeness of a nice API)  but have been concerned about the use of it already. A cursory search of the coi-services code base look like most places are not checking the return type and many are depending on it to throw an exception if there is a problem, which of course the current implementation is not.  The exception to this is that there are a few tests which are looking at the return value, so I might have to modify those.

So I guess I will go ahead and change the return type and the logging to exception raising; then merge.

PR #247, opened 2013-02-08
Adds hash_any to pyon

lukecampbell:


daf:
Two thoughts:
- since this doesn't depend on pyon, consider putting it in `putil/` which doesn't auto-gevent-monkey-patch
- any way to get the numpy import isolated into a smaller section so this is mostly functional without numpy?

daf:
Also if the Import fails, this method should still exist/return something.

blazetopher:
Dave - how's it looking now?

daf:
:shipit:

PR #254, opened 2013-02-12
Better sysname prepend to queue name, fixes OOIION-699

daf:


shenrie:
This seems to fix the problem I was seeing.

PR #270, opened 2013-03-05
Refactored Governance Controller to use only one Event Subscriber

shenrie:


daf:
Small quibble, but please use `"""` for docstrings in the future instead of `'''` - http://www.python.org/dev/peps/pep-0008/#documentation-strings

PR #274, opened 2013-03-08
Container management

newbrough:
event listener in container for management requests
update ion-defs and utilities versions


mmeisinger:
Looks good

PR #281, opened 2013-03-12
Refactor container modularity

mmeisinger:
DO NOT MERGE!

This PR addresses the following Jira tasks:
https://jira.oceanobservatories.org/tasks/browse/CIDEVCOI-651 (Modular capability container with varying capabilities)
https://jira.oceanobservatories.org/tasks/browse/CIDEVCOI-659 (Refactor datastore layer)

Based on: https://github.com/ooici/ion-definitions/pull/300

This PR will evolve based on feedback. Highlights:

Container profiles are explicitly declared in config files
Profiles can have selective container capabilities
Profiles can change the implementation of capabilities (keeping the use interface the same)
Profiles can change pyon.yml to adjust for that profile
Gumstix profile is now possible


pyon/container/procs.py
```diff
@@ -70,15 +70,16 @@ def start(self):
...
+            cc_obj = CapabilityContainer(name=self.container.id, cc_agent=self.container.name)
+            self.cc_id, _ = self.container.resource_registry.create(cc_obj)
 
-        #Create an association to an Org object if not the rot ION org and only if found
-        if CFG.container.org_name and CFG.container.org_name != CFG.system.root_org:
-            org, _ = self.container.resource_registry.find_resources(restype=RT.Org, name=CFG.container.org_name, id_only=True)
-            if org:
-                self.container.resource_registry.create_association(org[0], PRED.hasResource, self.cc_id)  # TODO - replace with proper association
+            #Create an association to an Org object if not the rot ION org and only if found
+            if CFG.container.org_name and CFG.container.org_name != CFG.system.root_org:
```
shenrie, line 79: This is kind of weird to allow some capability containers running that are not going to be associated with the Org in the RR. This may affect governance and seems to allow for inconsistencies in the resource data model.

pyon/container/procs.py
```diff
@@ -199,10 +201,9 @@ def spawn_process(self, name=None, module=None, cls=None, config=None, process_i
 
             else:
                 #Update local policies for the new process
-                if self.container.governance_controller:
+                if self.container.has_capability(self.container.CCAP.GOVERNANCE_CONTROLLER):
```
shenrie, line 204: This is ok as long as the has_capability(self.container.CCAP.GOVERNANCE_CONTROLLER) implies that the self.container.governance_controller is set and is not disabled.

lukecampbell:
Looking over the changes, I understand where things are going with the capability containers and the set oc capabilities. My only concern is ElasticSearch and how we tie that in to the new datastores (if we can).

daf:
Github's diff view has this all wrong for me - maybe the state of the master has changed and therefore is not showing me the changes here, it just looks like renaming some imports -> that's all.

mmeisinger:
This PR was manually merged and somehow the file diffs are screwed up.
Campfire still had the right diff:
https://github.com/ooici/pyon/compare/a8edca4af230...6e2d25778539

PR #298, opened 2013-03-25
ensure ionobj store unicode chars as utf8 encoded strings

pkediyal:
checks dictionaries and lists for unicode content and converts them to utf8 encoded strings

### DIFF
--- a/pyon/core/registry.py
+++ b/pyon/core/registry.py
@@ -148,6 +148,22 @@ class IonObjectRegistry(object):
 
                 if isinstance(value, unicode):
                     self.__dict__[name] = str(value.encode('utf8'))
+                elif isinstance(value, dict):
+                    for k, v in value.iteritems():
+                        if isinstance(k, unicode):
+                            del value[k]
+                            k = str(k.encode('utf8'))
+                            value[k] = v
+                        if isinstance(v, unicode):
+                            value[k] = str(v.encode('utf8'))
+                    self.__dict__[name] = value
+                elif isinstance(value, list):
+                    counter = 0
+                    for v in value:
+                        if isinstance(v, unicode):
+                            value[counter] = str(v.encode('utf8'))
+                        counter = counter + 1
+                    self.__dict__[name] = value
                 else:
                     self.__dict__[name] = value
 
@@ -172,4 +188,3 @@ class IonObjectRegistry(object):
             obj = clzz(**kwargs)
 
         return obj
-
--- a/pyon/core/registry.py
+++ b/pyon/core/registry.py
@@ -149,13 +149,19 @@ class IonObjectRegistry(object):
                 if isinstance(value, unicode):
                     self.__dict__[name] = str(value.encode('utf8'))
                 elif isinstance(value, dict):
+                    removeKey =[]
+                    addKeyValue={}
                     for k, v in value.iteritems():
                         if isinstance(k, unicode):
-                            del value[k]
+                            removeKey.append(k)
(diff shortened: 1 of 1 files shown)


daf:
This doesn't handle the case of recursing in a dictionary etc.  We have a walk method for this exact purpose: https://github.com/ooici/pyon/blob/master/pyon/core/object.py#L325-L361

See its use in the same file.  I'd like to see this redone to use walk instead.

Also, not sure if you can replace inline like you are doing?  I guess it does work if you're submitting this obviously but I thought python would complain about it.

pkediyal:
I did not see the problem you mention about python complaining, but I don't know why it should, please elaborate. 

If after reading what I say below you mention problem areas again I will evaluate once again.

The difference between what you propose and this is that walk is invoked only when couchdb does _ion_object_to_persistence_dict(), whereas this solution takes care of the conversion right when we create IonObject. Thus unicode is prevented from going over the messaging infrastructure.

I am not really sure of the overarching objective behind this all, but the JIRA issue mentions the following:
- the capability container with pyon.net and datastore is exclusively built using Python 2 and str types
- pyon.net can transmit objects with binary content and ASCII strings

I guess at the very least, encoding before putting messages on the wire will save some bandwidth as ascii will take 8 bits only but msgpack is doing that too anyway !.

daf:
My point was that you can call `walk` from this block of code and let it take care of iterating the dict/list for you, because it understands recursion.  You specify the callable (probably a lambda or small inline-def doing the str->unicode conversion if appropriate).

pkediyal:
OK .. that makes sense. I can try that out.

pkediyal:
Will it do lists though? It seems it only handles dicts.

daf:
It does lists.  Anything enumerable, in fact.  Just has to special-case dicts.

pkediyal:
Also it says  "Keys are left untouched". Whereas I am utf8-encoding those as well.

daf:
Ah, ok.  This may not directly apply then.  I just wanted to ensure traversing enumerables (dicts/lists) is done with common code to eliminate gotchas.  Maybe you can come up with something, or just leave as is.

pkediyal:
I would leave it as is.

However, I am still interested in your first comment regarding "you thought python will complain". Is there something I did not think through? Are you worried about infinite recursion like changing dicts inside setattr calling setattr, obviously this is not a problem but some problem like that in my solution? If you could elaborate I could think about it.

daf:
Sometimes when you change keys/entries of the very thing you are enumerating, next time you try to grab a value (in the for) it aborts with a "enumerable has changed" exception.  As long as you've tested it with more than one item, and it works ok, this is probably not that situation - probably happens more when adding/removing items rather than changing them in place.

mmeisinger:
Changing a collection (dict) while iterating over it must be avoided.
  http://stackoverflow.com/questions/6777485/modifying-a-python-dict-while-iterating-over-it

In particular changing the key could lead to a rehashing/redistributing of dict hash buckets. You change from unicode to str with utf8, which can change the has function in case a unicode content is really in. It will most certain fail in more general cases.

shenrie:
I am not comfortable with changing keys either. It seems like this might be going one step too far. It might be preferable to use the walk function Dave suggested and just be done with it.

---

System Architect

pkediyal:
So Michael, you did mention you wanted the keys changed as well. Do you still want me to attempt that with another solution than changing keys while iterating over the dict? Or just do what Dave and then Stephen mention?

PR #299, opened 2013-03-25
fixes problem with iterating over dictionary, ensures unicode to utf8-encoded strings

pkediyal:


pkediyal:
since Michael had requested that keys be utf-8 encoded as well, I fixed the problem due to modifying dictionary while iterating over them.

Inadvertently I closed the earlier pull requests (pyon and coi-service)

PR #300, opened 2013-03-26
Enhance list_queues to take optional column filtering

shenrie:


shenrie:
This column feature filtering enhancement could be extended to all of the management api queries, but that will have to come later

PR #301, opened 2013-03-27
Coi 655 - Ensure Unicode is stored as utf8 encoding

pkediyal:


mmeisinger:
The solution is ok.

(we assume a nested IonObject inside is just a dict that is already compliant - in theory you would not need to walk it anymore)

The coding is not PEP8 compliant. Please follow PEP8 when delivering production code. You can read about the rules here: https://confluence.oceanobservatories.org/display/CIDev/Code+Conventions and here: http://www.python.org/dev/peps/pep-0008/

I've seen the integration test in coi-services. A unit test in Pyon is required.

pkediyal:
fixed pep8 compliance added unit test

mmeisinger:
camelCase for functions and variables is not ok with PEP8. E.g. recursiveEncoding -> recursive_encoding; removeKey -> remove_key;                      addKeyValue -> add_key_value

pkediyal:
I ran PEP8 tool that didn't catch that, thanks for the education will put those changes in a minute

pkediyal:
this should be good to go now

pkediyal:
hope this does it !

PR #316, opened 2013-04-08
Fixes numpy encoding errors

lukecampbell:
I don't even know where to begin describing this bug.

Here is the outgoing granule before encoding

```
{'domain': (10,), 'data_producer_id': '', 'connection_index': '',
'record_dictionary': {1: array([ 0.,  1.,  2.,  3.,  4.,  5.,  6.,  7.,
8.,  9.]), 2: None, 3: None, 4: None, 5: None, 6: None, 7: None, 8:
None, 9: None, 10: None, 11: None, 12: None, 13: None, 14: None, 15:
None, 16: None, 17: None, 18: None, 19: None}, 'connection_id': '',
'locator': None, 'type_': 'Granule', 'param_dictionary':
'9a1f8b57e7af41479d42ce0764ad4ec1', 'creation_timestamp':
1365446562.004057, 'provider_metadata_update': {}}
```

Here it is after encoding

```
><> {'domain': [10], 'data_producer_id': '', 'connection_index': '',
>'record_dictionary': {1: [0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0,
>9.0], 2: None, 3: None, 4: None, 5: None, 6: None, 7: None, 8: None, 9:
>None, 10: None, 11: None, 12: None, 13: None, 14: None, 15: None, 16:
>None, 17: None, 18: None, 19: None}, 'connection_id': '', 'locator':
>None, 'type_': 'Granule', 'param_dictionary':
>'9a1f8b57e7af41479d42ce0764ad4ec1', 'creation_timestamp':
>1365446562.004057, 'provider_metadata_update': {}}
```

The numpy array was converted into a list and it turns out it stemmed
from the IonObject's overloading the setattr which is dynamically
attached at runtime inside registry.py and the walker bypasses walk and
somehow converts the numpy array into a list because numpy has **iter**.

Dave F suggested we use walk here so I implemented it using walk and now
granules flow again without bugs.

### DIFF
--- a/pyon/core/object.py
+++ b/pyon/core/object.py
@@ -332,15 +332,16 @@ def walk(o, cb):
 
     @TODO move to a general utils area?
     """
+    if _have_numpy and isinstance(o, np.ndarray):
+        return o
     newo = cb(o)
 
     # is now or is still an iterable? iterate it.
-    if _have_numpy:
-        if isinstance(newo, np.ndarray):
-            return newo
     if hasattr(newo, '__iter__'):
         if isinstance(newo, dict):
-            return dict(((k, walk(v, cb)) for k, v in newo.iteritems()))
+            return {k:walk(v,cb) for k,v in newo.iteritems()}
+        elif _have_numpy and isinstance(newo, np.ndarray):
+            return newo
         else:
             return [walk(x, cb) for x in newo]
 
--- a/pyon/core/registry.py
+++ b/pyon/core/registry.py
@@ -7,6 +7,7 @@ import inspect
 from copy import deepcopy
 
 from pyon.core.exception import NotFound
+from pyon.core.object import walk
 
 import interface.objects
 import interface.messages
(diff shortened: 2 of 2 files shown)


lukecampbell:
This bug is preventing the transport of granules correctly. And it is an extension of this PR:
https://github.com/ooici/pyon/pull/298

lukecampbell:
[This is where the culprit lives](https://github.com/ooici/pyon/blob/master/pyon/core/registry.py#L152)

`numpy` arrays extend `__iter__` and the `recusive_encoding` method converts the array in both the caller and the local object.

Here's an example of what's happening

When the deserializer iterates over the dictionary containing the granule it runs setattr(ion_obj, k,v) on all the field/values for the dictionary against the ion object, which in this case is a Granule.

The setattr is overridden and somehow coerces the numpy array into a list.

Can someone please take a look at the PR and make sure it's correct, I know based on an earlier PR that there was an issue with utf-8 keys and values but I'm out of the loop on the context; all I know is that until it's resolved we can't visualize some data products and I can't test a lot of components.

```
><> from interface.objects import Granule

><> g = Granule()

><> granule = rdt.to_granule()

><> vars(granule)['record_dictionary']
-->
{1: None,
 2: None,
 3: None,
 4: None,
 5: None,
 6: None,
 7: None,
 8: None,
 9: array([ 0.,  1.,  2.,  3.,  4.,  5.,  6.,  7.,  8.,  9.]),
 10: None,
 11: None,
 12: None,
 13: None,
 14: None,
 15: None,
 16: None,
 17: None,
 18: None,
 19: None}

><> for k,v in vars(granule).iteritems():
    setattr(g,k,v)
...

><> vars(g)['record_dictionary']
-->
{1: None,
 2: None,
 3: None,
 4: None,
 5: None,
 6: None,
 7: None,
 8: None,
 9: [0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0],
 10: None,
 11: None,
 12: None,
 13: None,
 14: None,
 15: None,
 16: None,
 17: None,
 18: None,
 19: None}

><> vars(granule)['record_dictionary']
-->
{1: None,
 2: None,
 3: None,
 4: None,
 5: None,
 6: None,
 7: None,
 8: None,
 9: [0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0],
 10: None,
 11: None,
 12: None,
 13: None,
 14: None,
 15: None,
 16: None,
 17: None,
 18: None,
 19: None}
```

pkediyal:
many issues for unicode to utf-8 here.

(1) a string would become None, because the utf8fix returns that for non-unicode
(2) does not convert dictionary keys to utf8 encoding
(3) bypasses numpy array, am not sure if numpy array's content also need to be utf-8 encoded.

lukecampbell:
if the contents of an array are unicode they MUST be left that way.  The numpy container must be unaltered in any way. No component in the system should be allowed to alter the content of the numpy array container.  I'm fine with doing whatever you guys want to fix this but as it stands right now granules are throwing out numpy arrays which is causing a ton of problems for me.

pkediyal:
Fully understand that Luke, we do need to fix it. The easiest fix is to revert to my solution only that we add an if condition that ignores numpy arrays (like in walk). But then we have a "walk" duplicate. So I am working on an enhanced walk that addresses the requirement from Michael regarding encoding keys in dicts as well (walk leaves dict keys untouched).

lukecampbell:
Potentially resolved by https://github.com/ooici/pyon/pull/319

PR #319, opened 2013-04-09
make "walk" function process dictionary keys; use it for utf-8 encoding

lukecampbell:


pkediyal:
I guess we should also change recursive encoding so that it only modifies lists and dicts. The way it (and "walk") is written right now all iterables that are not dicts will be converted to lists.

lukecampbell:
I think that's the right answer.

PR #320, opened 2013-04-12
modify walk to handle dicts and list instead of all __iter__ ables;

pkediyal:
Here is a thing Dave/Luke need to give some input on as well. The current solution is different from Luke's in a subtle way.

Earlier, in recursive_encoding the transformation occurred on the reference to the dict that had been passed in. Now, the transformation is of the dict. This means recursive_encoding changes by value not by reference (one has to pardon loose and ambiguous use of terms here because python passes by value but in case for dict the value is the reference). To unambiguously see what I mean notice the changes in the test_object.py included in the commit.

The solution isn't doing something incorrect, but a change is a change in behavior and in this case it happens due to the way walk behaves.

Our tests in coi-services and pyon and Luke's tests pass though, and I don't expect a problem.

### DIFF
--- a/pyon/core/object.py
+++ b/pyon/core/object.py
@@ -13,11 +13,6 @@ import StringIO
 from pyon.util.log import log
 from pyon.core.exception import BadRequest
 
-try:
-    import numpy as np
-    _have_numpy = True
-except ImportError as e:
-    _have_numpy = False
 
 built_in_attrs = set(['_id', '_rev', 'type_', 'blame_'])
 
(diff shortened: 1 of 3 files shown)


daf:
:thumbsup: Looks good to me

lukecampbell:
:metal: If the test passes then DM should be fine!

shenrie:
If the tests pass, then I will merge and we will hope for the best and fix if something is broke.

---

System Architect

lukecampbell:
I appreciate you guys thoroughly testing and verifying this, I know it's a pain but it'll save us a lot of time debugging.

PR #330, opened 2013-04-18
Change create_prepare_resource_support to not require a resource id and bump ion-defs

shenrie:
...bump ion-defs

### DIFF
--- a/pyon/net/transport.py
+++ b/pyon/net/transport.py
@@ -470,7 +470,8 @@ class AMQPTransport(BaseTransport):
         """
         Publishes a message on an exchange.
         """
-        log.debug("AMQPTransport.publish(%s): ex %s key %s", self._client.channel_number, exchange, routing_key)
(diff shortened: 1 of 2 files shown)


pyon/net/transport.py
```diff
@@ -470,7 +470,8 @@ def publish_impl(self, exchange, routing_key, body, properties, immediate=False,
         """
         Publishes a message on an exchange.
         """
-        log.debug("AMQPTransport.publish(%s): ex %s key %s", self._client.channel_number, exchange, routing_key)
+        #log.debug("AMQPTransport.publish(%s): ex %s key %s", self._client.channel_number, exchange, routing_key)
+        log.debug("AMQPTransport.publish(%s): ex %s key %s, size %d", self._client.channel_number, exchange, routing_key, len(repr(body))+len(repr(properties)))
```
daf, line 474: These `len(repr(` calls are going to be pretty expensive.  It should be the commented out one while the other one remains.

PR #334, opened 2013-04-23
Revert codec fix, endpoint now handles Exceptions instead of IonExceptions

daf:


daf:
Waiting on http://buildbot.oceanobservatories.org:8010/builders/pyon_test/builds/1312

daf:
Passed.

daf:
Need to test something else locally.

PR #360, opened 2013-05-15
Refined prepare resource support

shenrie:


pyon/ion/resource.py
```diff
@@ -883,7 +886,54 @@ def get_associated_resource_info(self, resource_id="", resource_type=None):
                 setattr(resource_data, 'resource_type', deco_value)
 
                 res_list,_ = self._rr.find_resources(restype=deco_value, id_only=False)
-                setattr(resource_data, field, res_list)
+
+                exclude_lcs_filter_value = resource_data.get_decorator_value(field, 'ExcludeLifecycleStates')
+                if exclude_lcs_filter_value is not None and exclude_lcs_filter_value.find(',') > -1:
```
daf, line 891: fyi could be simplified to `',' in exclude_lcs_filter_value`

daf, line 891: also does this if block mean you MUST specify a list of more than one exclude value?

shenrie, line 891: The above is not true... if exclude_lcs_filter_value is a None then the "in" will fail with Nonetype not iterable

daf, line 891: Sorry, meant that the second half could've been simplified (`in` is more 'pythonic' than using .find and comparing to -1).  Not a big deal obviously.

PR #362, opened 2013-05-15
Tune QoS on all amqp channels to prefetch 1 msg only

daf:
UNIT/INT tests pass.

This does not technically depend on ooici/ion-definitions#366 as that just provides a reference in the configuration.  The code will use a default value of 1 which is the intention.  It can be merged in any order.


shenrie:
Tim and/or Michael,

Are we to merge this or are we waiting until next week to do so?

---

System Architect

mmeisinger:
Looks good

PR #365, opened 2013-06-03
Validate that message size is below 20M

shenrie:


mmeisinger:
Looks good.

One point: Raising an exception may not be the way to indicate a dropped message in an interceptor. I'm not sure how we do it but we may need to set a flag in the invocation instread of just raising an exception. Exceptions are not expected in interceptors - it may still work for now

shenrie:
Currently all of the interceptors do this via exceptions and the end point code handles appropriately. It is too much to change for R2 at this point, but you might consider refactoring in R3 especially if refactoring the interceptor stack is on the table.

---

System Architect

mmeisinger:
no change requested - that's an item for the bucket list.

given it's consistent with other interceptors, it's good for R2

PR #370, opened 2013-06-05
Catch IonExceptions in outgoing interceptor stack on RPC response and react

daf:
Will wrap up the Exception with a custom message prepended to help understand flow.

```
/Users/asadeveloper/Documents/Dev/code/pyon/pyon/net/endpoint.py in _send(self, msg, headers, **kwargs)
    901             log.info("RPCRequestEndpointUnit received an error (%d): %s", res_headers['status_code'], res_headers['error_message'])
    902             ex = self.exception_factory.create_exception(res_headers["status_code"], res_headers["error_message"], stacks=stacks)
--> 903             raise ex
    904
    905         return res, res_headers

BadRequest: 400 - (while trying to send RPC response for conv-id ASA-DEVELOPERs-MacBook-Pro_local_40893-1) The message size 808 is larger than the max_message_size value of 256

><>
```

### DIFF
--- a/pyon/net/endpoint.py
+++ b/pyon/net/endpoint.py
@@ -1122,7 +1122,19 @@ class RPCResponseEndpointUnit(ResponseEndpointUnit):
         # sample (possibly) before we do any sending
         self._sample_request(response_headers['status_code'], response_headers['error_message'], msg, headers, result, response_headers)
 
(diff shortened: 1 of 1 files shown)


mmeisinger:
Looks good.

PR #371, opened 2013-06-05
Change filesystem to use sysname prefix

daf:
Will isolate test runs etc and prevent filesystem clean from wrecking an active container with datasets etc


lukecampbell:
Hey hey look at that it's finally getting integrated!  :yum:

PR #374, opened 2013-06-07
Ability to lock DotDicts to prevent accidental modifications of core 'constants'

ianfixes:
This patch adds a function to DotDict that allows you to lock its contents (stays in effect until you `clear()` it).  

This patch applies the functionality to PRED, RT, LCE, and so on.  If you try to overwrite their attributes, you get an error:

```
><> from pyon.public import RT

><> RT.InstrumentDevice
--> 'InstrumentDevice'

><> RT.InstrumentDevice = 3
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
/Users/iankatz/.virtualenvs/artoo/sdtkiai/coi-services/extern/pyon/scripts/pycc.pyc in <module>()
----> 1 RT.InstrumentDevice = 3

/Users/iankatz/.virtualenvs/artoo/sdtkiai/coi-services/extern/pyon/pyon/util/containers.pyc in __setattr__(self, key, value)
     89             raise AttributeError('%s conflicts with builtin.' % key)
     90         if dict.__getattribute__(self, DICT_LOCKING_ATTR):
---> 91             raise AttributeError('Setting %s on a locked DotDict' % key)
     92         if isinstance(value, dict):
     93             self[key] = DotDict(value)

AttributeError: Setting InstrumentDevice on a locked DotDict

><> 

```

Currently the existing container tests pass, and a new one has been added.

pyon tests are passing, with the exception of those that give HTTP errors.

### DIFF
--- a/pyon/ion/resource.py
+++ b/pyon/ion/resource.py
@@ -103,17 +103,20 @@ def load_definitions():
     ot_list.append('IonObjectBase')
     ObjectTypes.clear()
     ObjectTypes.update(zip(ot_list, ot_list))
+    ObjectTypes.lock()
 
     # Resource Types
     rt_list = getextends('Resource')
     rt_list.append('Resource')
     ResourceTypes.clear()
     ResourceTypes.update(zip(rt_list, rt_list))
+    ResourceTypes.lock()
 
     # Predicate Types
     pt_list = get_predicate_type_list()
     PredicateType.clear()
     PredicateType.update(zip(pt_list, pt_list))
+    PredicateType.lock()
 
     # Compound Associations
     get_compound_associations_list()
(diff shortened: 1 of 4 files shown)


lukecampbell:
Looks good to me :+1:

mmeisinger:
Looks good. pop is there twice. Can one be deleted or was this supposed to protect a different function?

ianfixes:
It was supposed to protect popitem, I must have hit paste too many times.

ianfixes:
Can this be merged?

mmeisinger:
I would prefer to leave the __locked__ key out of the __dict__ instead setting a bool for the unlocked case. In case where __locked__ exists (value is ignored), the DotDict is considered locked, if absent it's unlocked and like before.

This keeps the __dict__ clean for the majority of cases and is a true extension of behavior instead of a modification. DotDict is used quite a bit unlocked and it would be good to keep it as close to what it is now

ianfixes:
OK, that's all set.

mmeisinger:
:+1:

PR #376, opened 2013-06-17
gevent block plugin

jamie-cyber1:
Detects long running greenlets that block the os thread for more than 1 second in nosetests, similar to busy loop detect in R1.


mmeisinger:
:+1:

PR #384, opened 2013-06-25
Fix problem typos in DotDict's pop() and popitem() 

ianfixes:
This fixes a glaring error in DotDict where the return values of pop() and popitem() were not being returned.  

In the course of diagnosing this error, found a stray use of CFG.get instead of CFG.get_safe for a multi-level dict lookup.


ianfixes:
Bump.  Is there anything preventing this merge?

mmeisinger:
Looks good. Let's give it a try

PR #396, opened 2013-10-10
OOIION-1310 Adds function profiling utility to be used in conjunction with Breakpoint

lukecampbell:
Added time_profile which is a method annotation, it is excellent in identifying
the sources of slow components. Just annotate any methods you want to profile
for a test run with time_profile and the output is a neat indented profile
readout for the execution time of the method and the total time the program has
been running.

Part of [OOIION-1310](https://jira.oceanobservatories.org/tasks/browse/OOIION-1310)

### DIFF
--- a/pyon/util/breakpoint.py
+++ b/pyon/util/breakpoint.py
@@ -7,6 +7,8 @@
 '''
 import functools
 import traceback
+import time
+import sys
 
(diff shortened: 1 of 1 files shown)


lukecampbell:
@jamie-cyber1 Ready for review and merging.

PR #397, opened 2013-10-11
update installation instructions

pkediyal:
@jamie_cyber1 please review and merge


pkediyal:
jamie I recently tried to do installation from scratch (i.e. python bootstrap) and things got jumbled up until I realized that the README was probably incorrect. Please do make sure that the update I made are indeed required, otherwise reject this pull request.

Also, I am not sure if the numpy needs to be 1.7.1 but since it is so for coi-services I thought it might be needed here as well.

jamie-cyber1:
Excellent.

PR #399, opened 2013-10-16
Fixes pycc argument issue with broker management port

birdage:
The connection string assigned in pycc.py uses a static port number for the
admin utility, the latest version of RabbitMQ changes the port from 55672 to
15672, the pycc.py script should obey the pyon.local.yml config file in lieu of
a static value


birdage:
@lukecampbell please review

PR #406, opened 2013-11-08
OOIION-1358: Enabling the container to load agent config from object store /REL2.0

mmeisinger:


mmeisinger:
@jamie-cyber1 please review

PR #408, opened 2013-11-13
OOIION-1358: Unicode to str encoding applied in object store /REL2.0

mmeisinger:
- Added recursive unicode to str UTF-8 encoder
- Added unit test and performance test
- Some common object handing polish
- Make container object store a separate class over datastore
- Add int test for container object store


mmeisinger:
@jamie-cyber1 please review and merge

PR #411, opened 2013-11-20
OOIION-1262: Extended resource framework compound association can apply base type filter

mmeisinger:


mmeisinger:
@jamie-cyber1 please merge

PR #413, opened 2013-11-21
OOIION-1396: Less strict option for read_mult in datastore etc

mmeisinger:


mmeisinger:
@jamie-cyber1 please merge

PR #417, opened 2013-12-01
OOIION-1515: Extend procs.py to add default process info

mmeisinger:


mmeisinger:
@jamie-cyber1 please review

PR #419, opened 2013-12-02
Pin gevent version buildout

jamie-cyber1:


jamie-cyber1:
@mmeisinger please review.

PR #425, opened 2013-12-05
OOIION-1527: Agent restart for non-system restart case

mmeisinger:


mmeisinger:
@jamie-cyber1 please merge

PR #431, opened 2013-12-19
OOIION-1184: All event subscriber queues auto-delete by default

mmeisinger:


mmeisinger:
@jamie-cyber1 please merge

PR #436, opened 2014-01-06
OOIION-1653: Derive valid postgres database name for special sysnames

mmeisinger:


mmeisinger:
@jamie-cyber1 please merge - test exists

PR #446, opened 2014-01-08
OOIION-1662: Provide delete via update mechanism to postgres datastore

mmeisinger:


mmeisinger:
@jamie-cyber1 please merge

PR #449, opened 2014-01-23
Misc options to enable gevent block monitoring during pycc tests

jamie-cyber1:


jamie-cyber1:
@wbollenbacher please review and merge.

PR #451, opened 2014-01-28
OOIION-1697: find_associations should not return retired

mmeisinger:


mmeisinger:
@jamie-cyber1 please merge

PR #454, opened 2014-01-30
OOIION-1710: Set exchanges durable/non-auto-delete if configured

mmeisinger:


mmeisinger:
@jamie-cyber1 please review and merge

PR #456, opened 2014-02-05
Remove excessive debugging statements

jamie-cyber1:

PR #457, opened 2014-02-05
Remove tests looking for excessive logs, see #456

daf:


daf:
@jamie-cyber1, #456 removed some log statements that are actually tested (possibly overzealously).  This removes the offending tests/statements.

daf:
Stepping on Jamie's toes (see #458)

PR #459, opened 2014-02-05
Fix two broken tests due to remove of debug statements

jamie-cyber1:


jamie-cyber1:
@mmeisinger or @daf please review and merge.

daf:
:shipit:

PR #464, opened 2014-02-16
Patch int_test so not ALL os.kill gets mocked

jamie-cyber1:
Original test patch accidentally impacted code such as:
https://github.com/ooici/coi-services/blob/c24165f6cbce51ed5f358d20bb311224d873d28b/ion/agents/port/port_agent_process.py#L154


jamie-cyber1:
@daf please review and merge.

PR #465, opened 2014-02-16
Patch int_test so not ALL os.kill gets mocked

jamie-cyber1:


jamie-cyber1:
@daf please review and merge.

PR #468, opened 2014-02-18
OOIION-1753: Add pending channel numbers to Node for insulation

daf:
When requesting a channel number from Pika, when the correct set of circumstances leads to a channel not being fully allocated, our channel number selector wouldn't know about it but would cause Pika to crash, failing the container.

This adds a layer of insulation to ensure when a channel number is handed out, it won't be handed out again until it completes a round trip from Pika and gets wrapped by our AMQPTransport class.

This is to help solve the AMQCHAN IS NONE THIS SHOULD NEVER HAPPEN issues that occur from spurious greenlet interleaving with pika.

### DIFF
--- a/pyon/net/messaging.py
+++ b/pyon/net/messaging.py
@@ -183,6 +183,9 @@ class NodeB(BaseNode):
 
         transport = AMQPTransport(amq_chan)
 
+        # return the pending in collection (lets this number be assigned again later)
+        self.client._pending.remove(transport.channel_number)
+
         # by default, everything should have a prefetch count of 1 (configurable)
         # this can be overridden by the channel get_n related methods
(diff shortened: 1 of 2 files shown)


daf:
@jamie-cyber1 @mmeisinger Please review (vs master)

mmeisinger:
Please rename PR with Jira as prefix

mmeisinger:
:+1: Looks good. Can it be merged?

jamie-cyber1:
We still need to make sure tests don't fail on REL2.0.  If it helps the stability of test running, it should be merged.

jamie-cyber1:
Please document pull requests in jira, Dave.

PR #469, opened 2014-02-18
CIDEVSTC-99 - WKT handling by datastore/Postgres

brianmckenna:
M112 T133 - WKT handling by datastore/Postgres

Implements a WKT string search capability to datastore/Postgres (using PostGIS)

Handles all valid WKT string and the concept of a circle with radius as a point with a buffer to desired radius. Example {'wkt':'POINT(-70 40)', 'buffer':1.0} creates a circle centered on 40N,70W with a radius of 1.0 degrees.

### DIFF
--- a/pyon/datastore/datastore_query.py
+++ b/pyon/datastore/datastore_query.py
@@ -40,10 +40,9 @@ class DatastoreQueryConst(object):
     GOP_WITHIN_BBOX = GOP_PREFIX + "within"
     GOP_CONTAINS_BBOX = GOP_PREFIX + "contains"
 
-    WKT_PREFIX = "wkt:"
-    WKT_OVERLAPS = WKT_PREFIX + "overlaps"
-    WKT_WITHIN = WKT_PREFIX + "within"
-    WKT_CONTAINS = WKT_PREFIX + "contains"
+    GOP_OVERLAPS_GEOM = GOP_PREFIX + "overlaps_geom"
+    GOP_WITHIN_GEOM = GOP_PREFIX + "within_geom"
+    GOP_CONTAINS_GEOM = GOP_PREFIX + "contains_geom"
 
     ROP_PREFIX = "rop:"
     ROP_OVERLAPS_RANGE = ROP_PREFIX + "overlaps"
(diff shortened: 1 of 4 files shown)


brianmckenna:
@mmeisinger please review the implementation of WKT strings for geo search

includes https://github.com/ooici/coi-services/pull/1795

mmeisinger:
Very nice extension of the datastore query framework. :+1: :clap: Good tests too.

For consistency of the datastore query framework, I would prefer if the WKT operators were defined in the following way:
    GOP_OVERLAPS_GEOM = GOP_PREFIX + "overlaps_geom"  # Find objects with geometry overlapping given WKT geometry
    GOP_WITHIN_GEOM = GOP_PREFIX + "within_geom"      # Find objects with geometry within given WKT geometry
    GOP_CONTAINS_GEOM = GOP_PREFIX + "contains_geom"  # Find objects with geometry containing given WKT geometry

instead of adding a new type WKT_

and these functions:
    def overlaps_geom(self, col, geom_wkt):
    def contains_geom(self, col, geom_wkt):
    def within_geom(self, col, geom_wkt):

If this is too much effort to change now, we can wait for a subsequent refactoring (I can do that if needed).

mmeisinger:
Also please rename this PR and the other ones following the integration team's naming guidelines with a Jira ticket name as prefix

brianmckenna:
If we go with GOP_PREFIX we're going to need to add a conditional layer in the GOP_PREFIX branch

https://github.com/brianmckenna/pyon/blob/m112_t133/pyon/datastore/postgresql/pg_query.py#L80-L91

This branching in _build_where() seems to be what the *_PREFIX constants are designed for, no?

mmeisinger:
In terms of naming, I would like to keep all the geospatial operators with the GOP prefix, for better classification. You can either make an if-else inside to check for the suffix e.g. if op.endswith("_geom") or define a secondary prefix GOP:WKT:overlaps.

Also I'd prefer to name the operations by what they do (compare against a geometry vs. compare against a bbox) not in which representation they are in (WKT). So they should be named def overlaps_geom for instance.

Can you make these changes please?

brianmckenna:
I'll go with the endswith() check. I'll ping you and Maurice when ready for merge.

brianmckenna:
@mmeisinger @MauriceManning ready for review merge

mmeisinger:
Excellent and thanks for changing flexibly

PR #472, opened 2014-02-26
OOIION-1772: Postgres datastore update_doc_mult support _deleted

mmeisinger:


mmeisinger:
@daf or @jamie-cyber1 please merge

PR #473, opened 2014-02-26
OOIION 1386: Performance gains: move unicode handling out of validating_setattr

daf:
Now `validating_setattr` can safely be turned off.

We lose the ability to recursively transform unicode -> utf8 on set, such as in nested dicts/lists, but the need is not obvious - the boundary we care about most is in the IonObjectDeserializer, taking json from messaging or the datastore and turning it into IonObjects.

Roughly 25% speedup on its own, 50% speedup with `validating_setattr` disabled in configuration.


daf:
@mmeisinger please review (or @jamie-cyber1)

mmeisinger:
:+1:

PR #479, opened 2014-03-03
CIDEVSTC-194 : add type and persisted version info

pkediyal:
tests described in 
https://confluence.oceanobservatories.org/display/CIDev/M156+Schema+Versioning

caution: related ion-definition Pull Request https://github.com/ooici/ion-definitions/pull/506 needs to go in first

### DIFF
--- a/pyon/core/object.py
+++ b/pyon/core/object.py
@@ -499,26 +499,6 @@ class IonObjectDeserializer(IonObjectSerializationBase):
                 # in pyon metadata is in the document
                 # so we discard _attachments while transforming between the two
                 if  (v != None) and k not in ("type_", "_attachments", "_conflicts"):
-                    try:
-                        # hack: (ignore dicts)
-                        # if attribute is an ion object but doesn't have the same class
-                        if ( k not in built_in_attrs and ("type_" in v) and (ion_obj._schema[k]["type"] != v["type_"] )):
-                            # hack to ignore special case for tolerating current practice
-                            # of stuffing objects in dicts
-                            if (ion_obj._schema[k]["type"] == 'dict'):
-                                pass
-                            else:
-                                V = self._obj_registry.new(v["type_"])
-                                # check that v is subclass of k
-                                if not ion_obj.check_inheritance_chain(type(V), ion_obj._schema[k]["type"]):
-                                #if not issubclass(V.__class__,getattr(ion_obj,k).__class__):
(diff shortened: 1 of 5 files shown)


pkediyal:
A test case related to generating document fails when running -UNIT test. This is not serious in the sense that it won't break the system. But I will address it.

mmeisinger:
:+1: on the basic code changes and tests

@daf can you please briefly state whether the changes introduced here are compatible with your recent changes to unicode handling and object validation? Just looking for if these change do not cancel yours out.

pyon/core/object.py
```diff
@@ -479,7 +498,27 @@ def _transform(self, obj):
...
+                        # if attribute is an ion object but doesn't have the same class
+                        if ( k not in built_in_attrs and ("type_" in v) and (ion_obj._schema[k]["type"] != v["type_"] )):
+                            # hack to ignore special case for tolerating current practice
+                            # of stuffing objects in dicts
+                            if (ion_obj._schema[k]["type"] == 'dict'):
+                                pass
+                            else:
+                                V = self._obj_registry.new(v["type_"])
+                                # check that v is subclass of k
+                                if not ion_obj.check_inheritance_chain(type(V), ion_obj._schema[k]["type"]):
```
daf: Worried about the performance implications of these lines - if this does any iteration/walking of objects this may be very costly in the deserializer.  I can see why it is needed though and performance will always be a tradeoff.  Will get a chance to test more thoroughly later.

pkediyal: I guess this is handled at some point by the validator as well. But we don't really do anything about it except log an error. We can drop it once we run it by Tim, he was the last person who mentioned needing such a thing.

mmeisinger:
Production system performance is the #1 issue at this point. Anything that gets into a message's way must be lightning fast. No possibility for costly de/serializers

mmeisinger:
One thing to consider is that type version checking may only be required on load from the datastore, not for every de/serialization in messaging. How are these 2 cases treated differently?

daf:
We don’t have a specific configuration/derivation for each type of deserialization that happens.  Maybe that is the proper route here - have a IonObjectDatastoreDeserializer used in the datastore code that does this type of validation.

pkediyal:
so consensus is remove .. I will remove from deserializer. we can put it elsewhere later.

PR #481, opened 2014-03-05
M156: remove incorrect check for None values

pkediyal:


mmeisinger:
Looks good to fix the issue with None's at hand

PR #482, opened 2014-03-07
M156: add conditional persisted_version update during deserialization

pkediyal:
how about this? are there other databases being used? should I make similar changes for couchdb?


pkediyal:
OK. Nothing that a little bit of curry could not solve !

Good to go by me.

mmeisinger:
:+1:

PR #491, opened 2014-03-13
M058: OOI extern services integration to pyon

birdage:
M058:
Integration of externalization services to pyon, requires eoi.yml for use [pull request here](https://github.com/ooici/ion-definitions/pull/523). Allows EOI services to piggy page on resource CRUD requests. To actually leverage the services it requires that the [importerService](https://github.com/birdage/ooi-extern/blob/master/importerService.py) be setup and running [correctly](https://github.com/birdage/ooi-extern/blob/master/README.md).

see the following [link](https://confluence.oceanobservatories.org/display/CIDev/M058+OGC+web+services+suite+integration) for confluence page information.


mmeisinger:
Why is the code here not in the dataset management service, where Dataset resources are created?
I would like to keep the resource registry and Pyon clean from application level concerns like the one here.

Can we move this code into the dataset management service?

mmeisinger:
Also, please no CamelCase in the code such as useGeoServices, pyon.ion.tableLoader etc.

See https://confluence.oceanobservatories.org/display/CIDev/Code+Conventions

mmeisinger:
Also, if the importer service is OOI code, it needs to be in a GIT repo of the ooici account

birdage:
@mmeisinger ive outlined the replies to you comments below.
1) I dont know if we can move the code to the DMS, this would require a code review
2) Understood will refactor when possible.
3) We are in the process of moving birdage/ooi-extern to [here](https://github.com/ooici/ooi-extern)

mmeisinger:
Please move the code to DMS. Pyon/RR is not the place for it. We need to find a way to do so, even if it means fixing current code.

birdage:
@mmeisinger closing pull request for now.

PR #498, opened 2014-04-11
Adds asynchronous thread management R2

lukecampbell:
- Posix Thread and Gevent support


lukecampbell:
Thank you for not making a liar out me when I said "it'll be merged before the meeting is over".

PR #502, opened 2014-04-15
Thread Pool R2

lukecampbell:


lukecampbell:
Hold off on merging it just yet, I want to add some degree of synchronization for exiting the threads safely.

lukecampbell:
@mmeisinger ready for review

pyon/util/async.py
```diff
@@ -88,6 +87,62 @@ def get_pythread():
...
+    '''
+    libevent callback to read from the pipe
+    '''
+    try:
+        os.read(event.fd, 1)
+    except EnvironmentError:
+        pass # EAGAIN
+
+# create a libevent callback to read from the pipe
+gevent.core.event(gevent.core.EV_READ | gevent.core.EV_PERSIST,
```
mmeisinger, line 125: Is this a side effect from importing this module? Should this be marked? Does this have any consequences?

mmeisinger:
:+1: good job, looks good on read (without trying it out). Solid implementation

I'm a bit nervous because this is complex new concurrency logic with lots of potential issues to uncover, but let's try it. Anything you can make sure to flag warning conditions, check preconditions and assertions, etc would be helpful to make this as robust as we can expect.

How does this relate to the 
https://github.com/surfly/gevent/blob/master/gevent/_threading.py
module that Jeff would like to see? Would this be an extension here or replacing some of your code?

Did you take the thread pool base logic from somewhere? If yes, can you add a pointer so that we can follow it in case of questions

lukecampbell:
I got the basic concepts from threading.py and https://bitbucket.org/denis/gevent-playground/src/61bb12c9b4e41b58a763a7ef53fbe9a89cea1e04/geventutil/threadpool.py?at=default

I'll add links to them both in the code shortly. 

The problems with both is that they assume that all modules haven't been monkey patched. The threadpool.py example would need to be parsed and the threadpool initialized before any monkey patching, to work. I couldn't reliably get it to work on a container so I just went with a more primitive version that uses an un-monkey patched version of the built-in thread module.

pyon/util/async.py
lukecampbell, line 125: It is. Two new file descriptors will be created when the module is imported. I originally had the two file descriptors being created inside the class but as someone pointed out in a gevent forum about this, you quickly wind up with too many open files, and the gevent developer suggested this pattern. 

https://groups.google.com/forum/#!topic/gevent/tm8wNPZKoR8

lukecampbell, line 125: I can probably make an initializer for this, so that the file descriptors are created when they are needed first, not when the module is imported.

lukecampbell:
I'm gonna do some more work

PR #503, opened 2014-04-17
Gevent and Threading, working together

lukecampbell:


mmeisinger:
Please rename the module to threading, without underscore. Underscores are not a common name in pyon.

Other than that it looks like a job well done.

PR #505, opened 2014-04-17
Updates the threadpool to use the Queue from threading

lukecampbell:


mmeisinger:
(1) Looks good. But shouldn't all the code be in async or threading? There seems to be an artificial split between the two. I'd prefer if there were only one module and my first preference is async.

(2) I think we need to mirror all these PRs for R2.0

PR #507, opened 2014-05-01
Ensures force_clean doesn't run for pycc tests

lukecampbell:
There are tests that inherit from PyonTestCase but are part of the INT group of
integration tests. The PyonTestCase adds a file system cleanup at the
conclusion of the test runs. This patch does a quick env variable check to make
sure that a file system clean is allowed.


lukecampbell:
@jamie-cyber1 this should make the problems go away.

PR #511, opened 2014-05-21
CIDEVSTC-99 - M112 - support search radius in meters

brianmckenna:
If buffer value endswith('m'), the client has specified buffer in meters, WKT should be converted to its geographic representation (meters) and the buffer applied.

NOTE: the ST_ compare operators (contains, within, etc) require two geometry objects, so had to cast the buffered geography object back to geometry

@mmeisinger please review for merge

will push coi-services tests after merge and bump

### DIFF
--- a/pyon/datastore/postgresql/pg_query.py
+++ b/pyon/datastore/postgresql/pg_query.py
@@ -84,9 +84,9 @@ class PostgresQueryBuilder(object):
                 geom_from_wkt = 'ST_GeomFromEWKT(\'SRID=4326;%s\')' % (wkt)
                 # if buffer specified, wrap geometry in buffer http://postgis.net/docs/ST_Buffer.html
                 if buf:
-                    if buf.lower().endswith('km'):
+                    if buf.lower().endswith('m'):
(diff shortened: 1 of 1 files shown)


mmeisinger:
The impl is fine but why keeping the degree option around? Is there ever a need for users/callers to define a buffer by WGS84 degrees? In my limited understanding this would not even be a circle, because there are 360 longitude degrees vs 180 latitude degrees, resulting in an oval. Is there ever a science case where you would do this? Is it because it's relative to the coordinate system?
I would be ok with just moving to interpreting buf in meters (without suffix m) and leave it at this.
If you convince me we need both, the current impl is good.

brianmckenna:
The question of ever having a need should go in the other direction. I can discuss but i'm not a stakeholder.

My reasoning was as simple as our indexes are in WGS84, so the search circle was. I followed PostGIS' implementation of the buffer http://www.postgis.org/docs/ST_Buffer.html. I'm unaware of their reasons for having both vs. just meters.

Changes are trivial if needed, just let me know we're cleared for just meters and I'll make them.

mmeisinger:
let's just keep both for now. I'm merging and we can fix if needed as the UI team integrates.

PR #512, opened 2014-06-11
M153: Multiple Broker Support

daf:
Several underlying changes to pyon.net/pyon.ion.exchange infrastructure to more broadly support ExchangeObjects.  Endpoints automatically upgrade to using ExchangeObjects when the ability becomes available.  Events also updated.

PR for review / letting buildbot do it's thing.  I've tested rather extensively locally but as this is a very substantial, sweeping change, there are likely going to be ripples.


mmeisinger:
:+1: Very nicely done. I'm waiting for the builder results, but this is ready to go to pyon thereafter

